PR #90, opened 2012-01-31
Prevent crash while compiling extension-less files.

silentbicycle:
The default Rules file doesn't check if a file has an extension before attempting to append '.' to it. If you have an extension-less file in the content directory, it crashes.

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -64,7 +64,7 @@ end
 
 route '*' do
   if item.binary?
-    item.identifier.chop + '.' + item[:extension]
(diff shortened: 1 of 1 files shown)


oddityoverseer13:
I hate to bring up an old issue, but it seems this was never actually fixed. The change was made to the test code, but it's still wrong [here](https://github.com/nanoc/nanoc/blob/master/lib/nanoc/cli/commands/create-site.rb#L133)

ddfreyne:
@oddityoverseer13 Thanks for reporting! I’ve moved this into a new issue (#524).

PR #475, opened 2014-09-06
Added :commands_dirs (Array<String>) support to configuration file.

gpakosz:
Makes nanoc load commands from configurable directories.

This replaces the hardcoded `commands` path by `@config[:commands_dirs]`.


gpakosz:
Could this be viewed as an enhancement of #424 instead of a new feature? I'm afraid 3.8 won't be out any time soon.

ddfreyne:
I’m OK with releasing 3.8 soon, even if it doesn’t bring a lot of new stuff to the table. In that case, I’d aim for a release this weekend.

PR #477, opened 2014-09-06
add typohero filter

minad:
similar to typogruby etc, but faster and better


minad:
@ddfreyne Sure thx!

ddfreyne:
And it’s [released](https://rubygems.org/gems/nanoc-typohero)!

There’s some work to be done on granting everyone access to all nanoc-related gems, but that’s a separate issue.

PR #480, opened 2014-09-18
only upload file to AWS if MD5 hash has changed

paulboone:
Compares an MD5 hash of the local file to the stored MD5 on S3 (in the etag field) before uploading so that only changed files get uploaded to S3. Save time and $!

### DIFF
--- a/lib/nanoc/extra/deployers/fog.rb
+++ b/lib/nanoc/extra/deployers/fog.rb
@@ -73,19 +73,27 @@ module Nanoc::Extra::Deployers
       end
       keys_to_destroy = files.all.map { |file| file.key }
       keys_to_invalidate = []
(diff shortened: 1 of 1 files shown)


ddfreyne:
@mpapis Are you working on this now?

mpapis:
no I'm not, if you have time for it right now go for it, I probably wont have time before evening (it's waiting in my queue)

ddfreyne:
Let’s get this in 3.7.6!

ddfreyne:
I’ll take over this PR.

I’d like to have this change in 3.7.x rather than in master (3.8.0). This is not a feature and one could argue that uploading unchanged files is a bug.

ddfreyne:
Superseded by #536.

Issue #469, opened 2014-08-12
SystemStackError when updating from 3.6.9 to 3.6.10


ghost:
After updating I get:

`/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/forwardable.rb:174: stack level too deep (SystemStackError)`

Removed the _tmp_ directory and the error changes to:

`/Library/Ruby/Gems/2.0.0/gems/nanoc-3.6.10/lib/nanoc/base/checksummer.rb:25: stack level too deep (SystemStackError)`

Same happens with 3.7.1.

Is there anything I can do to narrow down the issue?

ddfreyne:
A fix for this, which allows items and layouts in attributes again, is in #518.

ddfreyne:
Fixed by #518. This’ll go in 3.7.5.

PR #507, opened 2014-12-25
Prevent items and layouts in attributes

ddfreyne:
This is a potential fix for #469.

This makes the checksummer explicitly abort when finding an item or a layout as an attribute, since this could potentially lead to an infinite loop, and is generally not well-supported.

An alternative fix could be to not allow items and layouts to end up in attributes in the first place.


ddfreyne:
Better fix in #518.

PR #515, opened 2015-01-01
Upgrade Bundler before install

ddfreyne:
This is necessary so that Travis uses the most recent Bundler version, 1.7.10, since that is what the gemspec requires.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ rvm:
   - 1.9.3
   - 1.8.7
   - rbx-2.2.6
+before_install: gem install bundler
 matrix:
   include:
     - rvm: jruby-1.7.11


ddfreyne:
The only build failure is the strange watcher error on JRuby.

mpapis:
we should not depend on `bundler` gem, it's supposed to handle dependencies, as such it can't be one.

ddfreyne:
Bundler is a development dependency though.

mpapis:
I still do not think bundler should be in Gemfile (/`.gemspec`), you could do the same things in new rubygems with:
```bash
gem i -g Gemfile
```
the only thing that would prevent us from continuing would be `Bundler` in `Gemfile`, but with this simple trick I got it going:
```patch
-all_rubies = Bundler::Dependency::PLATFORM_MAP.keys
+all_rubies = Bundler::Dependency::PLATFORM_MAP.keys rescue [:ruby_19, :ruby_20, :ruby_21]
```
there should be better code in rubygems to take care of it

ddfreyne:
Seems like we could get rid of Bundler entirely, and use just `gem install -g/--file`. I haven’t used that before, though, so I don’t know the drawbacks of doing it that way.

mpapis:
I had only one problem to fix manually (`nokogiri`), in the end this worked:
```bash
gem i nokogiri -- --use-system-libraries
gem i -g Gemfile
```
but `nokogiri` is not `bundler` or `rubygems` issue, it might just work on other systems without any extra steps

ddfreyne:
So what you’re suggesting is removing Bundler entirely. Right now, I’d rather resolve the problem that we’re seeing (fix in this PR), and perhaps look into removing Bundler in a future PR.

mpapis:
then I'm :+1: on this, will open new one to remove `bundler` dependency

ddfreyne:
OK, cool!

PR #518, opened 2015-01-10
Bound recursive checksumming

ddfreyne:
Fix for #469. Instead of preventing items and layouts in attributes, we can allow them and let the checksummer properly deal with it.

PR #507 had a different implementation, but too much like a hack.

### DIFF
--- a/lib/nanoc/base/checksummer.rb
+++ b/lib/nanoc/base/checksummer.rb
@@ -18,8 +18,13 @@ module Nanoc
 
       private
 
-      def update(obj, digest)
+      def update(obj, digest, visited = Set.new)
         digest.update(obj.class.to_s)
+
+        if visited.include?(obj)
+          digest.update('recur')
+          return
+        end
 
         case obj
         when String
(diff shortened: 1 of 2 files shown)


lib/nanoc/base/checksummer.rb
```diff
@@ -18,23 +18,28 @@ def calc(obj)
 
       private
 
-      def update(obj, digest)
+      def update(obj, digest, visited = Set.new)
         digest.update(obj.class.to_s)
 
+        if visited.include?(obj)
+          digest.update('recur')
```
mpapis: if visited would be a hash then we could store also the digest value and set it here, would it be useful at all?

ddfreyne: We can’t set the digest value because at the time it isn’t known yet.

mpapis: ah I see it now, painful

ddfreyne:
Can I get a +1?

mpapis:
looks good but the jruby test failed, I have restarted it to check if it was one time or an real problem

ddfreyne:
Tests are passing on Travis CI, but GitHub isn’t picking up the success. Odd!

PR #520, opened 2015-01-11
use older mustache for older rubies

mpapis:
mustache `1.0.0` requires ruby `2.0.0+`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,7 +23,11 @@ gem 'maruku'
 gem 'mime-types'
 gem 'minitest', '~> 4.0'
 gem 'mocha'
-gem 'mustache'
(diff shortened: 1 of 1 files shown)


ddfreyne:
I’d rather specify the 1.0.0+ version for Ruby 2.0:

```ruby
if RUBY_VERSION >= '2.0.0'
  gem 'mustache', '~> 1.0'
else
  gem 'mustache', '~> 0.99'
end
```

mpapis:
updated, waiting for tests, btw. this should be done against `release-3.7.x`?

ddfreyne:
Yup, this should be against `release-3.7.x`.

PR #522, opened 2015-01-27
Improve nanoc!

maciejczyzewski:
I made a few useful changes.

booyah

_I tested locally. It should work. (check me)_


mpapis:
It overall looks like to much to consider at once, like for sure you should separate adding RVM project files and un-hiding `autocompile`.

Also whats up with removal of new line at the end of few files, that would be problematic for anyone using shell tools to work with the files (like me using `grep` to find something).

mpapis:
now, can you separate this into separate PR's, some of the changes might be rejected and it influences the rest, also open the new PR's against `release-3.7.x` instead of `master`

maciejczyzewski:
There is no need... Everything we can examine here! :wink:

ddfreyne:
This PR has major issues, and provides no value.

Non-useful changes:

1. Rearranging _.gitignore_ is not useful.
2. Adding _.ruby-xyz` files is not useful.
3. Adding comments to _.travis.yml_ is not useful.
4. Changing the header of the README is not useful.
5. Adding installation instructions to the README is not useful, because nanoc already has [installation instructions](http://nanoc.ws/install/).
6. Changing the nanoc command description is not useful.
7. Rearranging the gemspec is not useful.

Breaking changes:

1. Explicitly setting `Encoding.default_internal` and `Encoding.default_external`.
2. Removing `Nanoc3`.
3. Un-deprecating the _autocompile_ command.
4. Un-deprecating the _watch_ command.

Interesting changes:

1. Adding `osx` to _.travis.yml_: interesting idea, but I’m not convinced this is useful. nanoc has no OS-dependent parts. If anything, it should add Windows. I’m more on the side of not adding OS X as a tested platform.

2. Updating the release branch name. Although, since it’s used as an example only, there’s not all that much value to changing it.

3. Adding `# encoding: utf-8`: perhaps useful, but then Rubocop config should be adjusted to take tasks and YARD handlers into account.

mpapis:
for `.ruby-version` - well adding some kind of ruby version specification helps those that use multiple rubies and would avoid installing nanoc dependencies in all the rubies, it's like **"recommended ruby version"**, I'm cursing lack of this file every time I switch to nanoc, I have to either find ruby I used last time or install all gems again, with `.ruby-*` files ruby switchers would automatically select proper ruby to work with ... or `.gitignore` all of them so users can define their own file.

bobthecow:
just add the file to the directory yourself. by forcing a version, you screw everyone else who doesn't want to use that version. it's actually kind of rude ;)

lifepillar:
Agree with @bobthecow. But having `.ruby-version` in `.gitignore` would be a good idea, in my opinion.

bobthecow:
Yeah, but not the end of the world even if it's not added. Did you know you can ignore things in your own copy of a repo, even if the project owner doesn't add them to `.gitignore`?

```bash
echo .ruby-version > .git/info/exclude
```

:)

ddfreyne:
I’ve recently switched to using the Ruby provided by Homebrew, but I could also have been using the default one provided by the OS.

There was a time where you almost *needed* to have multiple Rubies installed, but that time’s pretty much past.

bobthecow:
I still use rbenv because I've got enough production Ruby apps locked to various Ruby versions that I don't want to mess with keeping in sync :)

PR #532, opened 2015-02-15
Keep sorting order of articles passed to atom_feed.

th-h:
According to the documentation, atom_feed's :articles option
should default to sorted_articles, but in fact, it didn't.
Since 2012-09-24 (a9c947803185b63336b2a93184155f538fa32eba),
it defaulted to articles (unsorted )instead, and
relevant_articles would be sorted by :created_at in
descending order anyway (in sorted_relevant_articles).
Due to that, it was impossible to sort relevant_articles in
any other way (by updated_at or some other attribute).

This change restores the former behaviour without changing
the current default.

Signed-off-by: Thomas Hochstein <thh@inter.net>

### DIFF
--- a/lib/nanoc/helpers/blogging.rb
+++ b/lib/nanoc/helpers/blogging.rb
@@ -84,14 +84,12 @@ module Nanoc::Helpers
 
       protected
 
-      def sorted_relevant_articles
-        relevant_articles.sort_by do |a|
-          attribute_to_time(a[:created_at])
-        end.reverse.first(limit)
+      def limited_relevant_articles
+        relevant_articles.first(limit)
       end
 
       def last_article
-        sorted_relevant_articles.first
+        limited_relevant_articles.first
       end
 
       def validate_config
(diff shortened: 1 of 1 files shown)


bobthecow:
This is a backwards compatibility break for anyone relying on the current sorting behavior (i.e. just about anyone using this for anything but defaults). Maybe it should be fixed with another option to keep the current sorted order?

ddfreyne:
I think the bug is twofold:

* There is no ability to define a custom sort order.
* The documentation for `atom_feed` is incorrect, because it says the default for `:articles` is `sorted_articles` (should be `articles`).

As @bobthecow mentioned, solving this in a backwards-incompatible way requires more work. I was playing with the idea of having a `:sort_proc` option, but I imagine @bobthecow’s suggestion for an order-preserving option (e.g. `:preserve_order`) is simpler and easier to understand.

ddfreyne:
I created an issue for this: #533.

ddfreyne:
I’ll close this PR now, since it doesn’t solve the issue in a backwards-compatible way. Compatible implementations for #533 are definitely welcome though; if there are no takers, I’ll likely work on it tomorrow.

Issue #533, opened 2015-02-20
Sort order not preserved in Atom feed


ddfreyne:
The Atom feed helper *always* sorts by `created_at`. There is no way to change the sort order.

This is a bug, since it was previously possible to pass in an array and not have it sorted.

I propose a `preserve_order` option, defaulting to false (in order to stay backwards compatible with current behavior).

Also see #532.

bobthecow:
:+1:

ddfreyne:
Potential fix in #534.

th-h:
Yeah, I should have opened an issue before. Sorry.

My train of thought was, as the current behaviour (as of 3.7.5) already was a backwards compatibility break (introduced by a9c947803185b63336b2a93184155f538fa32eba in 2012), I could just fix it without adding another option (as the default behaviour without setting the :article option wouldn't change and everybody setting :article could just be sorted any way one likes).

I didn't bear in mind that I'd yet *again*  break backwards compatibility.

The fix in #534 works for me. Thank you!

PR #534, opened 2015-02-21
Add option to preserve order in Atom feed

ddfreyne:
Potential fix for #533.

This also fixes the incorrect default for the `:articles` option.


bobthecow:
:shipit:

Issue #526, opened 2015-01-31
Impossible to pass generic Pandoc options in Rules


ddfreyne:
It is impossible to pass these options to Pandoc in the Rules file:

```ruby
:s, {:f => :markdown, :to => :html}, 'no-wrap', :toc
```

ghiknt:
After further experiment it is also impossible to pass options twice (e.g. --filter).  Since the last one is all that is kept in the hash.  Temporally worked around by adapting test case to filter wrapper:

```ruby
class PandocCommandLine < Nanoc::Filter
  identifier :pandoc_commandline
  type :text
  def run(content, params={})
   filter = ::Nanoc::Filters::Pandoc.new
   filter.setup_and_run(content, *params[:args])
  end
end
```

Which is called by Rules file.
```ruby
opts = ['table-of-contents', {:to => :html5, 'base-header-level' => 2}, {:filter =>'pandoc-zotxt'}, {:filter => 'pandoc-citeproc'}]
filter :pandoc_commandline, :args => opts 
```

* Note table-of-contents is not a good one to use for testing though since it only gets rendered in latest pandoc when --standalone is specified.

ddfreyne:
Yeah, I feel that the only solution to this would be to introduce an arbitrary `:args` param, as you suggested. I think I’ll fix it like that; there’ll be a PR incoming soon.

ddfreyne:
Potential fix in #535.

ddfreyne:
Fixed by #535.

PR #535, opened 2015-02-21
Allow generic Pandoc options with :args

ddfreyne:
Potential fix for #526.

It is now possible to call Pandoc like this:

```ruby
filter :pandoc, args: [:s, {:f => :markdown, :to => :html}, 'no-wrap', :toc]
```

I don’t think this breaks backwards compatibility, even though the filter’s method signature has changed, because it was never possible to call the filter in a different way.

CC @ghiknt

### DIFF
--- a/lib/nanoc/filters/pandoc.rb
+++ b/lib/nanoc/filters/pandoc.rb
@@ -11,8 +11,10 @@ module Nanoc::Filters
     # @param [String] content The content to filter
     #
     # @return [String] The filtered content
-    def run(content, *params)
-      PandocRuby.convert(content, *params)
+    def run(content, params = {})
+      args = params.key?(:args) ? params[:args] : params
+
+      PandocRuby.convert(content, *args)
     end
   end
 end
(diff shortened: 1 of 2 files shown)


ddfreyne:
CC @nanoc/contributors — would like a +1

ghiknt:
Gladly.  I have manually patched against 3.7.5 using your diff and it works great.  Where do I +1?

ddfreyne:
Leaving a comment that says `:+1:` works!

I just realised that the documentation isn’t up-to-date. Will fix now.

ghiknt:
:+1:

ddfreyne:
The documentation is now updated. I’ve marked the `:args` approach as recommended, since it’s the most flexible and matches closely the way one would invoke Pandoc from the command line.

Are you OK with the documentation?

ddfreyne:
screen shot 2015-03-07 at 16 34 16

ghiknt:
Looks good to me.  My test/use case is:

```ruby
opts = [{:to => :html5,
        'base-header-level' => 2,
        :filter => 'pandoc-citeproc',
        'data-dir' => 'content/about/bibliography/',
        :bibliography => 'content/about/bibliography/whk.name.bib',
        :csl => 'content/about/bibliography/chicago-annotated-bibliography.csl'
       }, 'section-divs']
filter :pandoc, :args => opts  
```

Which I would have been able to develop from the new docs.

PR #536, opened 2015-02-21
Do not upload files with identical etags

ddfreyne:
This is continued off #480.

This is quite tricky to write tests for. Not sure how to handle that.

### DIFF
--- a/lib/nanoc/extra/deployers/fog.rb
+++ b/lib/nanoc/extra/deployers/fog.rb
@@ -63,6 +63,7 @@ module Nanoc::Extra::Deployers
         files += set
       end
       keys_to_destroy = files.all.map(&:key)
+      etags = read_etags(files)
 
       # Upload all the files in the output folder to the clouds
       puts 'Uploading local files'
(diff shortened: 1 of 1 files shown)


ddfreyne:
CC @nanoc/contributors — would like a +1

mpapis:
yes looks good :+1:, I guess my only remark is that the run grows quite big, but it's not related to this PR

ddfreyne:
Will look into adding some tests for this. (Will be hard, I guess!)

lib/nanoc/extra/deployers/fog.rb
```diff
@@ -70,11 +71,17 @@ def run
         files = Dir['**/*'].select { |f| File.file?(f) }
         files.each do |file_path|
           key = path ? File.join(path, file_path) : file_path
+
+          keys_to_destroy.delete(key)
+
+          local_etag = calc_local_etag(file_path)
+          remote_etag = etags[key]
+          next if remote_etag && remote_etag == local_etag
+
```
mpapis: avoid calculating local hash if remote is not available:
```ruby
remote_etag = etags[key]
if remote_etag
  local_etag = calc_local_etag(file_path)
  next if remote_etag == local_etag
end
```

mpapis:
to simplify testing you could extract https://github.com/nanoc/nanoc/blob/aws-deploy-md5-v2/lib/nanoc/extra/deployers/fog.rb#L68-L86 to separate method `upload` - this also solves problem of growing `run` - and add `attr_accessor :directory, :etags, :keys_to_destroy` - this way you can check your custom `directory` and list of `etags` without actually connecting to aws.

lib/nanoc/extra/deployers/fog.rb
ddfreyne: Thanks, updated! (see #552—not this PR)

Issue #537, opened 2015-02-23
Accessing pre snapshot in layout results in Circular dependency error


jethrogb:
When I try to access the :pre compilied snapshot for an item in a layout, I get a Circular dependency error. The final :pre snapshot gets saved at https://github.com/nanoc/nanoc/blob/7a297646/lib/nanoc/base/result_data/item_rep.rb#L380 just before the layout. Is there any reason I shouldn't be accessing this snapshot at this time?

Use case layout excerpt:
```html
<% if @item.compiled_content(snapshot: :pre).include? 'type="math/tex' then %>
<!-- include MathJax -->
<% end %>
```

ddfreyne:
Not being able to access the `:pre` snapshot seems like a bug to me.

First of all, to avoid the issue you’re seeing, you can use `yield` to get the content right before the layout is applied. For example:

```ruby
<% if yield.include? 'type="math/tex' then %>
```

[_item_rep.rb_, line 247](https://github.com/nanoc/nanoc/blob/7a297646/lib/nanoc/base/result_data/item_rep.rb#L247) has the check `!self.compiled? && is_moving`; if that expression is true, an unmet dependency error will be raised, resulting in the circular dependency issue.

That check is accurate for the `:post` and `:last` snapshot, since these snapshots keep moving until compilation is finished. For the `:pre` snapshot, this check can probably be relaxed, because `:pre` snapshots will not move again once they’re created.

So, instead of this:

```ruby
if @content[snapshot].nil? || (!self.compiled? && is_moving)
```

…it’d probably be possible to have this:

```ruby
if @content[snapshot].nil? || ([:post, :last].include?(snapshot) && !self.compiled?) || (snapshot == :pre && @content[:pre].nil?)
```

I don’t think there’s any reasonable situation where `yield` would not be usable, though.

ddfreyne:
On second thought, my proposed fix doesn’t work. This line:

```ruby
if @content[snapshot].nil? || ([:post, :last].include?(snapshot) && !self.compiled?) || (snapshot == :pre && @content[:pre].nil?)
```

is equivalent to

```ruby
if @content[snapshot].nil? || ([:post, :last].include?(snapshot) && !self.compiled?)
```

which clearly shows that `:pre` is not handled in a special way at all, even though it should, since it is a moving snapshot: just because a `:pre` snapshot exists, does not mean it cannot move in the future.

[...]

PR #538, opened 2015-03-07
Allow acessing :pre snapshot in item itself

ddfreyne:
Fix for #537.

The `:pre` snapshot can now be accessed within the item itself, provided that the `:post` snapshot is present (because at that point, the `:pre` snapshot stops moving).

The implementation of the fix in this PR is conceptually the same as in the aforementioned bug, but refactored.

CC @nanoc/contributors @jethrogb

### DIFF
--- a/lib/nanoc/base/result_data/item_rep.rb
+++ b/lib/nanoc/base/result_data/item_rep.rb
@@ -243,12 +243,20 @@ module Nanoc
         raise Nanoc::Errors::NoSuchSnapshot.new(self, snapshot)
       end
 
-      # Require compilation
-      if @content[snapshot].nil? || (!self.compiled? && is_moving)
+      # Verify snapshot is usable
+      is_still_moving =
+        case snapshot
+        when :post, :last
+          true
+        when :pre
+          !@content.key?(:post)
+        end
+      is_usable_snapshot = @content[snapshot] && (self.compiled? || !is_still_moving)
+      unless is_usable_snapshot
         raise Nanoc::Errors::UnmetDependency.new(self)
-      else
-        @content[snapshot]
       end
+
+      @content[snapshot]
     end
 
     # Checks whether content exists at a given snapshot.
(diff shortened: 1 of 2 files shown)


ddfreyne:
Can I get a code review?

mpapis:
:+1: looks reasonable - although I do not have full understanding of it so I might be missing something

bobthecow:
Is the `:pre` snapshot usable after other (explicit) snapshots have been created but before `:post` is present?

ddfreyne:
Yeah, `:pre` is always usable.

bobthecow:
Then maybe I missed something? I thought this was setting it up to allow accessing `:pre` once `:post` is present?

ddfreyne:
@bobthecow This allows accessing `:pre` _from the item itself_, once `:post` is set. Previously, accessing `:pre` was only possible from other items.

bobthecow:
Right. So I was asking if you should be able to access `:pre` _from the item itself_, once, e.g., `:mysnapshot` is set.

ddfreyne:
@bobthecow Hm, re-reading the original question, I think my answer is misleading.

From within the item itself, using `:pre` when there is no `:post` snapshot will lead to an exception (since you’d recursively let compiled content depend on itself). When there is a `:post` snapshot, that works from within the item itself (since `:pre` cannot move anymore).

When accessing the compiled content from another item, accessing `:pre` is possible even if there is no `:post`: nanoc will compile the accessed item first, guaranteeing that `:pre` is in its final position.

These moving pre-snapshots and post-snapshots are difficult to reason about and one of the reasons why they’ll be gone in nanoc 4!

gpakosz:
Does it make `:pre` accessible at the time you're laying out an item? A while ago, when I tried to extract a TOC from `@item.compiled_content(:snapshot => :pre)` inside my default `layout.haml` layout, I think nanoc bailed out. So I used `yield` instead:

```haml
%main#content.container
  - toc = extract_toc(yield, :depth => @item[:toc_depth])
```

ddfreyne:
@gpakosz That’s still not possible, but `yield` is intended to be used in that case anyway.

PR #540, opened 2015-04-04
Removes specific version of cri.

froderik:
...nanoc complained about the 2.3 version of cri and bundler seemed to use 2.6 anyway. Removing the version works fine. (For me...)

UPDATE
Gets this warning:

    WARN: Unresolved specs during Gem::Specification.reset:
        cri (~> 2.3)
    WARN: Clearing out unresolved specs.
    Please report a bug if this causes problems.

Seems to me that cri is not used at this point. I don't understand why.

### DIFF
--- a/nanoc.gemspec
+++ b/nanoc.gemspec
@@ -27,7 +27,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 1.9.3'
 
-  s.add_runtime_dependency('cri', '~> 2.3')
+  s.add_runtime_dependency('cri')
 
   s.add_development_dependency('bundler', '>= 1.7.10', '< 2.0')
 end


bobthecow:
`~> 2.3` includes 2.6.

froderik:
Yep - it puzzles me. I updated the comment with the warning I get.

gpakosz:
I've always had this warnings. Typically, upgrading `nanoc` upgrades `cri` and then you have 2 different versions of `cri` and you get the warning. Uninstalling the older one solves it. But it did it every single time the scenario mentioned above happened.

froderik:
Aha - that explains it. Does it need to have av version in the gemspec?

ddfreyne:
I’ve seen this issue multiple times, and not just with nanoc. The cause is unclear; some people suggest `gem cleanup cri` or `bundle clean --force`.

Are you running with `bundle exec`? Do you get the issue with `bundle exec` too?

In any case, I’d say that the specific version of cri is correct; nanoc requires cri `~> 2.3`.

ddfreyne:
I’ll close this PR for now, because it’s not the proper solution.

Issue #542, opened 2015-04-16
New Check Proposal: Link protocol


jugglinmike:
We've recently purchased an SSL certificate for serving our nanoc-powered site. I'd like to avoid mixed content warnings by ensuring all embedded content is specified with a secure protocol or a protocol-relative URL. In order to prevent regressions, I'd like to implement an automated test for this.

This seems in line with nanoc's current checks--would a patch implementing `nanoc check secure-links` (or similar) be considered?

ddfreyne:
A good idea!

This shouldn’t be too difficult to implement. Off the top of my head, I think the following changes will need to be made:

* The existing links collector will probably need to be updated to capture some context (`img src`, `a href`, `link href`, …) so that it can restrict itself to included resources only.

* The base URL will need to be set in the site configuration so that the checker knows whether a URL is acceptable.

ddfreyne:
You can take a stab at implementing this if you want!

PR #543, opened 2015-04-18
Implement "Mixed Content" check

jugglinmike:
Commit message:

> This check ensures that resources embedded within HTML documents are not
> requested using a protocol that degrades the security of the original
> connection. This avoids "mixed content" warnings in browsers that
> enforce consistent protocols, and it precludes the leaking of private
> information in browsers that do not.

Resolves gh-542

### DIFF
--- a/lib/nanoc/extra/checking/checks/mixed_content.rb
+++ b/lib/nanoc/extra/checking/checks/mixed_content.rb
@@ -1,0 +1,31 @@
+# encoding: utf-8
+
+module Nanoc::Extra::Checking::Checks
+  # A check that verifies HTML files do not reference external resources with
+  # URLs that would trigger "mixed content" warnings.
+  class MixedContent < ::Nanoc::Extra::Checking::Check
+    PROTOCOL_PATTERN = /^(\w+):\/\//
+
+    def run
+      filenames = output_filenames.select { |f| File.extname(f) == '.html' }
+      resource_uris_with_filenames = ::Nanoc::Extra::LinkCollector.new(filenames).filenames_per_resource_uri
+
+      resource_uris_with_filenames.each_pair do |uri, fns|
+        next unless guaranteed_insecure?(uri)
+        fns.each do |filename|
+          add_issue(
+            "mixed content include: #{uri}",
+            subject: filename)
+        end
+      end
+    end
+
+    private
+
+    def guaranteed_insecure?(href)
+      protocol = PROTOCOL_PATTERN.match(href)
+
+      protocol && protocol[1].downcase == 'http'
+    end
+  end
+end
--- a/lib/nanoc/extra/link_collector.rb
+++ b/lib/nanoc/extra/link_collector.rb
@@ -4,6 +4,17 @@ require 'set'
 
 module ::Nanoc::Extra
   class LinkCollector
+    URI_ATTRS = {
+      'a' => :href,
+      'audio' => :src,
+      'form' => :action,
+      'iframe' => :src,
+      'img' => :src,
+      'link' => :href,
+      'script' => :src,
+      'video' => :src
+    }
+
     def initialize(filenames, mode = nil)
       Nanoc::Extra::JRubyNokogiriWarner.check_and_warn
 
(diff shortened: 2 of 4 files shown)


lib/nanoc/extra/link_collector.rb
```diff
@@ -4,6 +4,18 @@
 
 module ::Nanoc::Extra
   class LinkCollector
+
+    @@uri_attrs = {
```
ddfreyne: I’d prefer to have this in a constant rather than a class variable (I think Rubocop will complain anyway).

ddfreyne:
This looks good! I’ll do a more in-depth review soon, but I‘m quite liking it so far.

jugglinmike:
Great, thanks! I'll hold off on moving that class variable till you get a chance to finish.

lib/nanoc/extra/link_collector.rb
```diff
@@ -4,6 +4,18 @@
...
+      'a' => :href,
```
ddfreyne: Does it make sense to disallow linking to a non-secure site from a secure site?

lib/nanoc/extra/link_collector.rb
```diff
@@ -33,24 +45,46 @@ def filenames_per_href
...
+      filenames_per_resource_uri
+    end
+
     def external_href?(href)
       href =~ %r{^(\/\/|[a-z\-]+:)}
     end
 
-    def hrefs_in_file(filename)
-      hrefs_in_file = Set.new
+    def uris_in_file(filename, tag_names)
```
ddfreyne: Renaming this method breaks backwards compatibility of the API. Could you change it back, or let `hrefs_in_file` proxy to the new method?

(I’m not sure whether `LinkChecker` should’ve been part of the public API, but it’s too late now anyway!)

ddfreyne: Nevermind, just saw that you already did that. Sweet :)

ddfreyne: (FYI: you can mark classes/methods/modules/… as not part of the public API using `@api private`.)

lib/nanoc/extra/checking/checks/mixed_content.rb
```diff
@@ -0,0 +1,31 @@
...
+        next unless guaranteed_insecure?(uri)
+        fns.each do |filename|
+          add_issue(
+            "mixed content include: #{uri}",
+            subject: filename)
+        end
+      end
+    end
+
+    protected
```
ddfreyne: I go for `private` only these days as I find `protected` to be not so useful.

lib/nanoc/extra/checking/checks/mixed_content.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    @@protocol_pattern = /^(\w+):\/\//
```
ddfreyne: Same comment about the constant vs class variable as below.

ddfreyne: (I think you might have to use `@private` to hide it in the YARD documentation.)

ddfreyne:
The style checks are failing. Apparently, they’re also failing on other parts of the code—I’ll fix it; ignore the style test results for now.

lib/nanoc/extra/link_collector.rb
jugglinmike: 1. Prevent insertion of insecure "active content" (such as JavaScript) that could generate misleading content or steal private information
2. Prevent passive leaking of information--HTTP requests are sent in-the-clear and might leak information about the user's browsing session

Users who navigate using insecure hyperlinks are susceptible to `#2`, but navigation is performed actively by the user, and it's harder to achieve the task of "saving users from themselves". Plus, the current site's maintainer has no control over the availability of SSL on remote servers.

ddfreyne:
Can you rebase onto master? That should fix the style issues (at least the ones that are not your fault, heh).

ddfreyne:
Looks good otherwise!

jugglinmike:
Alrighty, I've rebased and incorporated your feedback in separate commits. I'm happy to squash when you think this is ready

lib/nanoc/extra/checking/checks/mixed_content.rb
```diff
@@ -0,0 +1,31 @@
+# encoding: utf-8
+
+PROTOCOL_PATTERN = /^(\w+):\/\//
```
ddfreyne: Could you move this into the `MixedContent` class? This way, `PROTOCOL_PATTERN` won’t be a top-level constant. Constant lookup will still work.

lib/nanoc/extra/link_collector.rb
```diff
@@ -2,6 +2,17 @@
 
 require 'set'
 
+URI_ATTRS = {
```
ddfreyne: Same as above (move into LinkCollector).

lib/nanoc/extra/checking/checks/mixed_content.rb
jugglinmike: Ah yeah, much better that way.

ddfreyne:
Cool! OK to squash. I’ll properly test it out tomorrow and will merge it if it’s OK!

jugglinmike:
Excellent--all squashed

ddfreyne:
Thanks!

jugglinmike:
My pleasure :)

PR #544, opened 2015-04-18
Add Vagrant integration

jugglinmike:
I don't take it for granted that the maintainers would even *want* to commit to this technology, but since I had already created this configuration for myself (in support of gh-543), I figured it couldn't hurt to share it (just in case).

Commit message:

> Vagrant [1] (together with VirtualBox [2]) provides a consistent and
> reproducable interface for creating a development environment. This
> reduces the "set up" documentation required by this project, supports
> contributors who work in a wide range of environments, and minimizes
> friction when the project's requirements change.
>
> [1] https://www.vagrantup.com/
> [2] https://www.virtualbox.org/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ doc/yardoc
 /.yardoc
 *~
 /Gemfile.lock
+/.vagrant
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,18 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure(2) do |config|
+  config.vm.box = "ubuntu/trusty64"
+
+  config.vm.provider "virtualbox" do |v|
+    v.memory = 1024
+  end
+
+  config.vm.provision "shell", inline: <<-SHELL
+    sudo apt-get update
+    sudo apt-get install -y build-essential zlib1g-dev ruby-dev gem
+    sudo gem install bundler
+    cd /vagrant
+    bundle install
+  SHELL
+end


gpakosz:
I'm not sure this should be merged. For example, the choice of Ubuntu Trusty is arbitrary.

Maybe as an example to nanoc.ws? On the contrary, if it gets merged, this could help people wanting to contribute to nanoc from Windows

jugglinmike:
The guest OS is definitely arbitrary. Trusty is a widely-deployed, stable, free, and open source option, which seemed like the most relevant formal criteria we could apply.

I think you're right about Windows users--they stand the most to gain from this. But even on Linux, I have a hard time with tools like rbenv and rvm.

I understand your reluctance to accept this, as it represents a commitment to maintain another piece of infrastructure. The reason it may be preferable to documentation is that it is machine-executable, so issues of "staleness" and developer error are less prominent.

mpapis:
This is similar to travelling ruby version - we should decide on something
that will simplify flow for development.

ddfreyne:
I’m in general not opposed to merging this, but I also feel it doesn’t belong in the root. Perhaps a _tools_ dir, with _Vagrantfile.sample_ in it? This way, it would be more clear that using Vagrant is not the recommended way, while still giving the opportunity to use it with a premade Vagrantfile.

FYI, I got rid of rvm/rbenv/chruby entirely and now only use Homebrew-installed Ruby. So much simpler.

DaveEveritt:
Just my opinion: I think it's great for the community to share dev setups, and although it's not really the job of nanoc to proscribe a particular solution, it would be good to have a 'place' to share them (nanoc website?). Also, Vagrant has a VM requirement, which many users won't need.

bobthecow:
It seems like this is the wrong place for a Vagrantfile... it should be in a project, not in the framework the project uses, right?

jugglinmike:
My motivation here was not to suggest a single preferred way to *consume* nanoc. I agree with the recent comments that such a recommendation would be inappropriate for inclusion in the project's source tree (but that it could work as part of the general user documentation).

Instead, this would be a way to *develop* nanoc. The goal here would be to make it easier for potential contributors to get involved in the project. If that was decided to be something in need of improvement, introducing a Vagrantfile (even one that is somewhat opinionated) into the project itself would make sense.

ddfreyne:
@jugglinmike I think there are too many options for developing Ruby apps. Some would suggest rvm, others rbenv, others chruby, others Homebrew’s Ruby, others system Ruby, others several Vagrant images with different Rubies, etc.

I suppose this Vagrantfile applies to most Ruby projects, not necessarily just nanoc. Because of this, I feel that the Vagrantfile doesn’t belong here.

I’ll close this issue for now, but I’m definitely interested in following up.

jugglinmike:
Sure thing. I appreciate the careful consideration that went into this. One important detail that makes this distinct from a general-purpose configuration for any Ruby project is the `build-essential` and `zlib1g-dev` packages--some minor gotchas to compiling some of nanoc's gem dependencies with native code.

Another approach would be to document these details, although I suppose in some ways that is the job of the other gems themselves... At the same time, it kind be kind of disruptive for new contributors to run `bundle install` and then have to search the web to resolve the problem(s).

PR #547, opened 2015-05-01
Fix double snapshot creation error

ddfreyne:
The following does not raise an error, but should:

```ruby
snapshot(:foo, stuff: :giraffe)
snapshot(:foo, stuff: :donkey)
```

The following does raise an error, but should not:

```ruby
snapshot(:foo)
snapshot(:bar)
```

The reason is that uniqueness is tested on the *second* argument (params), but should be on the first (name).


gpakosz:
:+1:

PR #548, opened 2015-05-01
Allow accessing :pre snapshot inside layout

ddfreyne:
Second part of the fix for #537.

This PR has two commits:

1. The fix itself, which consists of recording the `:pre` snapshot in the `snapshots` instance variable (which is filled in only with data from explicit snapshots), and let `#compiled_content` use the snapshots variable to determine whether the `:pre` snapshot is usable.

2. Some cleanup (e.g. renaming `snapshot` to `snapshot_name` to avoid confusion).

The code related to the fix isn’t as nice as it could be:

* Ideally, I’d redo the `snapshots` instance variable to be a hash instead of an array (not sure why it is an array in the first place) but that breaks backwards compatibility in the API. :(

* The `snapshots` instance variable is a bit weird anyway. It is used to check whether `#compiled_content` is trying to access a snapshot that exists, but I believe `@content` can be used for that purpose.

* I would also love snapshots to be a first-class concept. `Nanoc::Snapshot` would be great—instead of `snapshot[0]`, we’d have `snapshot.name`, and instead of `snapshot[1]`, we’d have `snapshot.final?`.

### DIFF
--- a/lib/nanoc/base/result_data/item_rep.rb
+++ b/lib/nanoc/base/result_data/item_rep.rb
@@ -235,29 +235,29 @@ module Nanoc
       Nanoc::NotificationCenter.post(:visit_ended,   item)
 
       # Get name of last pre-layout snapshot
-      snapshot = params.fetch(:snapshot) { @content[:pre] ? :pre : :last }
-      is_moving = [:pre, :post, :last].include?(snapshot)
+      snapshot_name = params.fetch(:snapshot) { @content[:pre] ? :pre : :last }
+      is_moving = [:pre, :post, :last].include?(snapshot_name)
 
       # Check existance of snapshot
-      real_snapshot = snapshots.find { |s| s.first == snapshot }
-      if !is_moving && (real_snapshot.nil? || real_snapshot[-1] == false)
-        raise Nanoc::Errors::NoSuchSnapshot.new(self, snapshot)
+      snapshot = snapshots.find { |s| s.first == snapshot_name }
+      if !is_moving && (snapshot.nil? || snapshot[-1] == false)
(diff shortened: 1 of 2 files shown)


ddfreyne:
@jethrogb This is the fix you’re looking for, I believe.

ddfreyne:
CC @nanoc/contributors — would be great to have a review, so this can end up in the upcoming release!

jethrogb:
Thanks for this PR. Works perfectly for me.

jethrogb:
> The `snapshots` instance variable is a bit weird anyway. It is used to check whether `#compiled_content` is trying to access a snapshot that exists, but I believe `@content` can be used for that purpose.

Quite weird, I feel like there's something missing. I grepped through the code, and it looks like the only uses of this variable are https://github.com/nanoc/nanoc/blob/release-3.7.x/lib/nanoc/base/result_data/item_rep.rb#L242 and 
https://github.com/nanoc/nanoc/blob/release-3.7.x/lib/nanoc/base/result_data/item_rep.rb#L435 and
https://github.com/nanoc/nanoc/blob/release-3.7.x/lib/nanoc/base/compilation/compiler.rb#L330 . In particular, manually calling ItemRep.snapshot does not change this variable. Is this wanted behavior?

ddfreyne:
@jethrogb Yeah. It is confusing, I have to admit.

The `@snapshots` instance var is filled in by `Nanoc::Compiler#compile_reps`. It runs through the _Rules_ file, detects all snapshots and stores them inside `Nanoc::ItemRep` instances. The names of the snapshots are used solely to detect whether the snapshot given to `#compiled_content` is one that exists.

This is definitely a candidate for refactoring!

PR #549, opened 2015-05-01
Allows creation of site in current directory

TheLonelyGhost:
Had issues when moving from Jekyll to nanoc with Github Pages where it didn't properly detect if there was a site that existed already.

I'm open to suggestions where detecting the presence of 'nanoc.yaml' assumes a site exists.

### DIFF
--- a/lib/nanoc/cli/commands/create-site.rb
+++ b/lib/nanoc/cli/commands/create-site.rb
@@ -4,7 +4,7 @@ usage 'create-site [options] path'
 aliases :create_site, :cs
 summary 'create a site'
 description "
-Create a new site at the given path (defaults to current working directory). The site will use the `filesystem_unified` data source by default, but this can be changed using the `--datasource` command-line option.
+Create a new site at the given path. The site will use the `filesystem_unified` data source by default, but this can be changed using the `--datasource` command-line option.
 "
 
 module Nanoc::CLI::Commands
@@ -280,7 +280,7 @@ EOS
       data_source = options[:datasource] || 'filesystem_unified'
 
       # Check whether site exists
-      if File.exist?(File.join(path, 'nanoc.yaml'))
+      if File.exist?(path) && (!File.directory?(path) || !(Dir.entries(path) - %w{ . .. }).empty?)
         raise Nanoc::Int::Errors::GenericTrivial, "A site at '#{path}' already exists."
       end
 
--- a/lib/nanoc/cli/commands/create-site.rb
+++ b/lib/nanoc/cli/commands/create-site.rb
@@ -4,7 +4,7 @@ usage 'create-site [options] path'
 aliases :create_site, :cs
 summary 'create a site'
 description "
-Create a new site at the given path. The site will use the `filesystem_unified` data source by default, but this can be changed using the `--datasource` command-line option.
+Create a new site at the given path (defaults to current working directory). The site will use the `filesystem_unified` data source by default, but this can be changed using the `--datasource` command-line option.
 "
 
 module Nanoc::CLI::Commands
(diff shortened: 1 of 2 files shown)


TheLonelyGhost:
Ah shoot, should I have requested this be merged into a different branch instead?

ddfreyne:
Could you elaborate on the issue you had where nanoc didn’t properly detect a site already existed? It should error out if the given destination directory already exists (even if it’s a file instead of a directory).

I’m uncertain whether this enhancement is useful, for two reasons:

* The current behavior seems clear to me, even if it doesn’t match the behavior of git. There’s not much additional value to defaulting to the current directory.

* The current behavior does not match the behavior of git on purpose: creating a site in the current directory might create a lot of new files, potentially overwriting some, and undoing that is a pain.

TheLonelyGhost:
You're right that defaulting to the current directory is less clear than the current documentation, but it addresses my problem:

I have a project that I want documented using gh-pages. It currently uses Jekyll on a gh-pages branch and I want to replace it to use nanoc.

Circumstances:

- I have a project hosted locally in `~/myproject` containing a super awesome Rust package that I am now circling back to update documentation.
- Project docs are being kept in a separate branch of the repository (`gh-pages`), which only contains Rust project source code necessary for documentation.
- Current docs use Jekyll to render them as a small website using Github Pages, but I want to use nanoc instead of Jekyll.
- I create a fresh branch (`docs`) for the nanoc, which will contain the nanoc source files.
- I will output the generated assets for the nanoc site to the `gh-pages` branch, completely replacing the Jekyll source files.
- The current working directory is `~/myproject` and the currently checked-out branch is `doc`

Given the above, I have 2 options:

1. cd into `~/myproject` and create a new site in the current directory
  - Not currently supported without this PR
2. cd into my home directory and run `nanoc create-site myproject`
  - Fails saying a site already exists (even though empty directory is specified)
  - This PR addresses the issue by detecting if `nanoc.yaml` exists in the specified directory instead

ddfreyne:
Being able to create a site into an empty directory (which fits your use case) seems useful. I’d change the `if File.exist?(File.join(path, 'nanoc.yaml'))` check (formerly `if File.exist?(path)`) to be a little safer, though. I recommend checking whether either the file does not exist at all, or is an empty directory:

```ruby
unless !File.exist?(path) || (File.directory?(path) && Dir.entries(path) == ['.', '..'])
  # …
end
```

Also, I’m afraid I’ve done some work on the create-site command and your changes are not longer mergeable. Could you rebase onto `master`?

ddfreyne:
Slightly more readable:

```ruby
if File.exist?(path) && (!File.directory?(path) || Dir.entries(path) != ['.', '..'])
  # …
end
```

TheLonelyGhost:
Rebasing now. I'm removing the sections where it defaults to the current directory since you make valid points.

On the other hand, checking if the directory is empty may not be right response either since my use case assumes a fresh git branch. Even a fresh branch will not be empty since it has to contain the `.git` directory. I hardly think hardcoding use-cases for all version control systems would be a good option either.

Would a prompt requiring confirmation be better if the directory is not empty?

ddfreyne:
If you mean an interactive prompt, then I’d say “no”, simply because there are no interactive parts in the nanoc CLI elsewhere.

Printing a warning and aborting when the given directory exists and is not empty makes sense. I’d say that message could say this:

> The given directory, _(full path here)_, exists and is not empty. If this is truly where you want the nanoc site to be created, re-run this command with `--force`.

Also, a new `--force` that overrides the non-empty-directory check.

TheLonelyGhost:
Now that master has stable builds again, looks like tests will pass. :+1:

I've limited the current directory functionality PR to this one and split off the `--force` option discussed above to #579. I'll let you choose if one or both are merged in.

ddfreyne:
:+1:

ddfreyne:
Build on master is failing but that’s unrelated to this PR (and I can’t even reproduce it locally…)

TheLonelyGhost:
Seems that Travis-CI has [had issues in the past week](https://www.traviscistatus.com/) so it might be related to that. Either way, the final commit that was merged had a successful build so it truly is unrelated to this PR.

PR #552, opened 2015-05-04
Do not upload files with identical etags

ddfreyne:
This is continued off #480. Replaces #536 (code identical, but this time on `master` instead of `release-3.7.x`).

This PR has quite a bit of extra tests and refactorings that make these tests possible. I also tested this with an actual S3 and can confirm that it works.


ddfreyne:
Got a +1 before the extra tests, so I’m assuming that’s fine now too. I tested it and it works anyway :)

PR #563, opened 2015-05-09
Rename internal monkey-patched methods

ddfreyne:
This will allow us to get rid of the monkeypatched methods, and explicitly break when the old names are used.


bobthecow:
:+1:

PR #568, opened 2015-05-10
Glob patterns

ddfreyne:
This adds glob patterns to nanoc 4. They are not super useful yet, because identifiers are still old-style (without extensions), but it is a good step forward.

Also, I guess I lied about nanoc 4 not getting new features!

Note that `File::FNM_EXTGLOB` is only available in Ruby 2.0 and up. Because of this, nanoc now depends on Ruby ≥ 2.0.0. (This should be fine, since Ruby 1.9.3 reached end-of-life status last February.)

Documentation for this new feature:

## Glob patterns in Rules

nanoc 4 supports using globs. Globs are more powerful than nanoc’s original pattern syntax, and they are also more commonplace, such as in Unix shells.

To enable globs, set `pattern_syntax` to `"glob"` in the configuration. For example:

```yaml
pattern_syntax: "glob"
```

The three most useful wildcards are the following:

`*`
: Matches any file or directory name. Does not cross directory boundaries. For example, `/projects/*.md` matches `/projects/nanoc.md`, but not `/projects/cri.adoc` nor `/projects/nanoc/about.md`.

`**`
: Matches any file or directory name, and crosses directory boundaries. For example, `/projects/**/*.md` matches both `/projects/nanoc.md` and `/projects/nanoc/history.md`.

`[abc]`
: Matches any single character in the set. For example, `/people/[kt]im.md` matches only `/people/kim.md` and `/people/tim.md`.

`{foo,bar}`
: Matches either string in the comma-separated list. More than two strings are possible. For example, `/c{at,ub,ount}s.txt` matches `/cats.txt`, `/cubs.txt` and `/counts.txt`.

nanoc 4 uses Ruby’s [`File.fnmatch` method](http://ruby-doc.org/core/File.html#method-c-fnmatch) with the `File::FNM_PATHNAME` option enabled. Please consult the [`File.fnmatch`](http://ruby-doc.org/core/File.html#method-c-fnmatch) documentation for other supported patterns, and more comprehensive documentation.

Patterns based on regular expressions are still supported in nanoc 4, so you can still use e.g. `%r{\A/projects/(cri|nanoc)\.md\Z}` to match both `/projects/nanoc.md` and `/projects/cri.md`.

### DIFF
--- a/lib/nanoc/base.rb
+++ b/lib/nanoc/base.rb
@@ -27,6 +27,8 @@ end
 
 # @api private
 module Nanoc::Int
+  require_relative 'base/pattern'
+
   # Load helper classes
   autoload 'Context',              'nanoc/base/context'
   autoload 'Checksummer',          'nanoc/base/checksummer'
--- a/lib/nanoc/base/compilation/compiler_dsl.rb
+++ b/lib/nanoc/base/compilation/compiler_dsl.rb
@@ -254,11 +254,20 @@ module Nanoc::Int
       @rules_collection.parse(filename)
     end
 
-    private
-
+    # @api private
     def create_pattern(arg)
-      Nanoc::Int::Pattern.from(identifier_to_regex(arg))
+      case @config[:pattern_syntax]
+      when 'glob'
+        Nanoc::Int::Pattern.from(arg)
+      when nil
+        Nanoc::Int::Pattern.from(identifier_to_regex(arg))
+      else
+        raise Nanoc::Int::Errors::GenericTrivial,
+          "Invalid pattern_syntax: #{@config[:pattern_syntax]}"
+      end
     end
+
+    private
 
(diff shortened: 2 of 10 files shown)


bobthecow:
I'd say `{foo,bar}` is one of the more useful glob patterns for a static site generator. So it's prolly worth mentioning here?

ddfreyne:
Good point. I added this:

> **`{foo,bar}`**
> Matches either string in the comma-separated list. More than two strings are possible. For example, `/c{at,ub,ount}s.txt` matches `/cats.txt`, `/cubs.txt` and `/counts.txt`.

ddfreyne:
Tests pass locally. I’ll merge; a +1 afterwards is fine too. There’s some more work that I’d like to do and it depends on this!

PR #569, opened 2015-05-10
Full style identifiers

ddfreyne:
When `identifier_style` is set to `"full"` in the data source config, identifiers with extensions are returned.

This also removes the static data source, which at this point has no use anymore whatsoever.

* [x] Add documentation
* [x] Add tests
* [x] Allow using globs to specify layout
* [x] Allow using globs to find items
* [x] Figure out what to do with child/parent relationship
* [x] Use full-style identifiers by default
  * [x] Set `pattern_syntax` to `"glob"` in default configuration
  * [x] Set `identifier_style` to `"full"` in default configuration
  * [x] Update default Rules file

Notes:

* Items with full-style identifiers do not have children or parents, because the child-parent relationship is ambiguous in these cases.

Example configuration:

```yaml
data_sources:
  -
    type: filesystem_unified
    identifier_style: full
```

Example new _Rules_ file that is now usable:

```ruby
compile '/**/*.sass' do
  filter :sass
end

compile '/**/*.md' do
  filter :kramdown
  layout '/default.html'
end

compile '/**/*.html' do
  layout '/default.html'
end

compile '/**/*' do
end

route '/**/*.md' do
  item.identifier.with_ext('html')
end

route '/**/*.sass' do
  item.identifier.with_ext('css')
end

route '/**/*' do
  item.identifier.to_s
end

layout '/**/*', :erb
```


ddfreyne:
Works like a charm for nanoc.ws (see nanoc/nanoc.ws#90).

bobthecow:
yay globs!

ddfreyne:
As for child-parent relationships: I think I’ll disable those for new-style identifiers. They don’t make sense anymore; the parent is not necessarily unique, and children are ambiguous.

```
/foo.md
/foo.html
/foo/donkey.md
/foo/giraffe.md
```

The parent of both `/foo/donkey.md` and `/foo/giraffe.md` could be `/foo.md` or `/foo.html`, and similarly, the children of `/foo.md` and `/foo.html` could both be `/foo/donkey.md` and `/foo/giraffe.md`.

bobthecow:
:+1: sounds good to me.

Issue #570, opened 2015-05-10
View API lacks useful methods


ddfreyne:
* `ConfigView`
  * `#fetch` (like `Hash#fetch`)
  * `#key?` (like `Hash#key?`)
* `ItemView`
  * `#fetch` (like `Hash#fetch`)
  * `#key?` (like `Hash#key?`)
  * `#parent` (like `Nanoc::Int::Item#parent`)
* `MutableItemView`
  * `#update_attributes` to bulk update attributes
* `MutableItemCollectionView`
  * `#concat` to bulk add items

Other related changes:

* `items.create` must return nil, so that it does not expose the internal entity.

ddfreyne:
Idea: `AttributesView` that wraps attributes.

cdlm:
Idea: a mixin that implements the rest of the `Hash` protocol, when you need objects that are like hashes.

ddfreyne:
See fix in #572.

PR #572, opened 2015-05-11
View API convenience methods

ddfreyne:
Fixes for #570.

Two ugly things:

* The dependency tracking support. Having to evaluate `@item[key]` just so the dependency tracking doesn’t break is pretty icky.

* The amount of duplication. I couldn’t find a clean and non-intrusive way to implement this.

I’m okay with both ugly bits, because they’re abstracted away and we can refactor this later. (I’d *love* to refactor the dependency tracking out of `Nanoc::Int::Item` and into the view classes.)

### DIFF
--- a/lib/nanoc/base/views/config.rb
+++ b/lib/nanoc/base/views/config.rb
@@ -3,6 +3,9 @@
 module Nanoc
   class ConfigView
     # @api private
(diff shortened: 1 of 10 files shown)


ddfreyne:
Merging this! You can give me a :+1: later.

PR #573, opened 2015-05-11
Document + fix view API

ddfreyne:
This adds documentation to the view API where it was missing or incomplete.

**Also:**

- I added specs where necessary and fixed some issues—all discovered while writing documentation.

- I removed `#concat`, because it doesn’t make sense (there is no way to get a collection of items that isn’t already in the site).


ddfreyne:
Again, you can give me a :+1: later :)

PR #574, opened 2015-05-12
Fix #passthrough

ddfreyne:
Discovered through nanoc/nanoc.ws#90.


gpakosz:
Does this bug affect 3.8?

ddfreyne:
Nope. It only affects 4.x.

gpakosz:
:+1:

PR #576, opened 2015-05-13
Fix rendering helper

ddfreyne:
This lets `#render` take a glob as its argument, e.g. `render '/page.*'`

For this to work, there’s now `LayoutCollectionView#[]`. Unfortunately, this makes it clear that the collection views are starting to show a serious issue of code duplication.


ddfreyne:
I’m quite tempted to not do anything about the code duplication for the time being, since it’d likely end up refactoring `Nanoc::Int::ItemArray`, which I’d like to avoid as long as 3.x is still the main stable version.

PR #580, opened 2015-05-13
Adds --force option to create site in non-empty directory

TheLonelyGhost:
Continuation of #549. Allows --force flag to override the site-already-exists safeguard

### DIFF
--- a/lib/nanoc/cli/commands/create-site.rb
+++ b/lib/nanoc/cli/commands/create-site.rb
@@ -282,7 +282,9 @@ EOS
 
       # Check whether site exists
       if File.exist?(path) && (!File.directory?(path) || !(Dir.entries(path) - %w{ . .. }).empty?) && !options[:force]
-        raise Nanoc::Int::Errors::GenericTrivial, "A site at '#{path}' already exists. Use --force to override this warning."
+        raise Nanoc::Int::Errors::GenericTrivial,
+          "The site was not created because '#{path}' already exists. " +
+          "Re-run the command using --force to create the site anyway."
       end
 
       # Check whether data source exists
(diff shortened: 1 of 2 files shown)


lib/nanoc/cli/commands/create-site.rb
```diff
@@ -281,7 +282,7 @@ def run
 
       # Check whether site exists
       if File.exist?(path)
```
ddfreyne: Can you move the `unless options[:force]` next to the condition here?

```ruby
if File.exist?(path) && !options[:force]
```

The `options[:force]` will otherwise be overlooked too easily, I feel.

lib/nanoc/cli/commands/create-site.rb
```diff
@@ -281,7 +282,7 @@ def run
...
-        raise Nanoc::Int::Errors::GenericTrivial, "A site at '#{path}' already exists."
```
ddfreyne: The error message isn’t as clear as it could be. Suggestion:

> The site was not created, because #{path} already exists. If you really want to create a site in this directory, re-run this command with `--force`.

ddfreyne:
Looks nice so far!

For an even better user experience, you could check whether `path` is a directory, and if it is, allow a site to be created with `--force`; it if is not a directory, don’t allow the site to be created (it would fail anyway).

ddfreyne:
You’ll have to rebase on top of master; the other PR conflicts with this one.

TheLonelyGhost:
It was originally based off of the other PR, but I figured it would be better kept as independent features in case one was accepted and the other rejected.

Rebased from master and moved the `!options[:force]` condition as per your suggestion while addressing merge conflicts.

ddfreyne:
All good except for the message itself. “A site already exists” is incorrect. I propose the following:

> The site was not created, because #{path} already exists. If you really want to create a site in this directory, re-run this command with `--force`.

What do you think?

TheLonelyGhost:
I feel like your suggestion is a bit wordy, but I can't think of an alternative that gives the same feel as the other messages in a more concise manner. Copypasta verbatim into the new commit message.

ddfreyne:
Thanks!

PR #581, opened 2015-05-14
Attributes entity

ddfreyne:
This is an experiment for using `Nanoc::Int::Attributes` instead of `Hash`.


ddfreyne:
This isn’t working out as well as I had expected. Mustache integration is making it difficult, too.

ddfreyne:
Will revisit this some other time.

Issue #589, opened 2015-05-17
Auto-prune is not enabled by default


ddfreyne:
It makes sense for nanoc to manage the contents of the _output/_ directory. In order to do so properly, it needs to have `auto_prune` turned on by default.

```yaml
prune:
  auto_prune: true
```

This will make sure that after every compilation, all stray files in the _output/_ directory are removed.

ddfreyne:
Fixed by #590.

bobthecow:
:+1:

PR #590, opened 2015-05-17
Auto-prune by default

ddfreyne:
This enables auto-pruning by default.

Fix for #589.


bobthecow:
:+1:

bobthecow:
might be worth calling this out in the config, or in the site setup, or in a file in the output folder. we don't want to surprise anyone by deleting their only copy of something.

Issue #588, opened 2015-05-16
Default Rules file is not suited for clean URLs


ddfreyne:
The Rules file is not well-suited for clean URLs. A file `content/about.html` will be routed to `output/about.html` rather than `output/about/index.html`.

For the record, the default Rules file in nanoc 4 currently looks like this:

```ruby
#!/usr/bin/env ruby

compile '/**/*.html' do
  filter :erb
  layout '/default.*'
end

compile '/**/*' do
end

route '/**/*' do
  item.identifier.to_s
end

layout '/**/*', :erb
```

ddfreyne:
Here’s an alternative Rules file that is more extensible. Important changes:

* It generates clean URLs for HTML files.

* It also includes a `.rdoc` extension, because I believe it is a good idea to introduce the concept of filtering by extension, and RDoc is part of Ruby anyway, so there are no external dependencies.

* It handles /index.* files properly.

```ruby
compile '/**/*.html' do
  layout '/default.*'
end

compile '/**/*.rdoc' do
  filter :rdoc
  layout '/default.*'
end

route '/**/*.{html,rdoc}' do
  case item.identifier
  when '/index.*'
    '/index.html'
  else
    item.identifier.without_ext + '/index.html'
  end
end

route '/**/*' do
  item.identifier.to_s
end

layout '/**/*', :erb
```

bobthecow:
I'd probably put a rule (commented out) for handling `*.md`, just because that's likely to be the most common question.

TheLonelyGhost:
:+1: for `*.md` rule in default Rules file, though that's likely off-topic for this issue.

Perhaps I'm behind on the history of nanoc, but why was this feature implemented in the first place? Is the appearance of `/about/` in the browser's navigation bar really cleaner than `/about.html`? What about the effect it has on cluttering the filesystem with directories holding exactly one file?

Both options seem clean to me and the latter indicates the format on the actual URL while making no assumptions about the index file on the HTTP server.

bobthecow:
http://www.w3.org/Provider/Style/URI.html

ddfreyne:
I prefer `http://nanoc.ws/about/` over `http://nanoc.ws/about.html`. The former is easier to read, and also easier to remember/type.

With nanoc’s routing rules, it’s still pretty easy to *not* use this approach (some people prefer `about.html`).

ddfreyne:
Potential fix in #591.

ddfreyne:
Fixed by #591.

PR #591, opened 2015-05-17
Better default site

ddfreyne:
The new default site now includes a sample compilation rule for RDoc, along with a sample RDoc file (hopefully clearly enough marked as “just for demonstration purposes”). It also comes with a commented-out compilation rule for Markdown files.

It also adds `rdoc` to the list of text extensions, because otherwise the site wouldn’t compile :)

Potential fix for #588.

### DIFF
--- a/lib/nanoc/base/source_data/site.rb
+++ b/lib/nanoc/base/source_data/site.rb
@@ -33,7 +33,7 @@ module Nanoc::Int
     # that lacks some options, the default value will be taken from
     # `DEFAULT_CONFIG`.
     DEFAULT_CONFIG = {
-      text_extensions: %w( css erb haml htm html js less markdown md php rb sass scss txt xhtml xml coffee hb handlebars mustache ms slim ).sort,
+      text_extensions: %w( css erb haml htm html js less markdown md php rb sass scss txt xhtml xml coffee hb handlebars mustache ms slim rdoc ).sort,
       lib_dirs: %w( lib ),
       commands_dirs: %w( commands ),
       output_dir: 'output',
--- a/lib/nanoc/cli/commands/create-site.rb
+++ b/lib/nanoc/cli/commands/create-site.rb
@@ -107,11 +107,27 @@ EOS
 #!/usr/bin/env ruby
 
 compile '/**/*.html' do
-  filter :erb
   layout '/default.*'
 end
 
+# This is an example rule that matches Markdown (.md) files, and filters them
+# using the :kramdown filter. It is commented out by default, because kramdown
+# is not bundled with nanoc or Ruby.
+#
+#compile '/**/*.md' do
+#  filter :kramdown
+#  layout '/default.*'
+#end
+
 compile '/**/*' do
+end
+
+route '/**/*.{html,md}' do
+  if item.identifier =~ '/index.*'
+    '/index.html'
+  else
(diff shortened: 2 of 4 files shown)


ddfreyne:
I also got carried away and [simplified file writing](581b7cca59a4d3899405eb66a64f3cbf6421df49).

lib/nanoc/cli/commands/create-site.rb
```diff
@@ -107,13 +107,36 @@ def array_to_yaml(array)
 #!/usr/bin/env ruby
 
 compile '/**/*.html' do
-  filter :erb
```
bobthecow: :+1: on not using an erb filter by default :)

ddfreyne: Yeah, that was weird.

ddfreyne:
I think the comment for the RDoc rule should be altered somewhat. It is currently this:

> This is an example rule that matches RDoc (.rdoc) files, and filters them
>  using the :rdoc filter. A newly created site comes with one RDoc-formatted
> file, content/samples.rdoc, for demonstration purposes.

I believe this will be more clear:

> This is an example rule that matches RDoc (.rdoc) files, and filters them
>  using the :rdoc filter. A newly created site comes with one RDoc-formatted
> file, content/samples.rdoc, for demonstration purposes. If you do not need RDoc,
> feel free to remove this rule as well as the content/samples.rdoc file.

bobthecow:
i would use parenthesis or em dashes instead of commas around `content/samples.rdoc`.

ddfreyne:
Oh, it also needs to tell to replace `{html,rdoc,md}` with just `{html,md}`.

> This is an example rule that matches RDoc (.rdoc) files, and filters them
>  using the :rdoc filter. For demonstration purposes, a newly created site
> comes with one RDoc-formatted file (content/samples.rdoc).
>
> If you do not need RDoc, feel free to remove this rule, the content/samples.rdoc file,
> and replace `{html,rdoc,md}` with just `{html,md}`.

That’s getting pretty wordy. :(

Having thought about this some more, I think I would rather *not* have anything RDoc-like in there by default. Nobody uses it (I know, because nobody complained about `rdoc` not being in text_extensions). Everyone uses Markdown instead.

bobthecow:
So just having a commented out markdown rule should be enough, yeah?

ddfreyne:
Yeah, I think I’ll do it that way.

lib/nanoc/base/source_data/site.rb
```diff
@@ -33,7 +33,7 @@ class Site
     # that lacks some options, the default value will be taken from
     # `DEFAULT_CONFIG`.
     DEFAULT_CONFIG = {
-      text_extensions: %w( css erb haml htm html js less markdown md php rb sass scss txt xhtml xml coffee hb handlebars mustache ms slim ).sort,
+      text_extensions: %w( css erb haml htm html js less markdown md php rb sass scss txt xhtml xml coffee hb handlebars mustache ms slim rdoc ).sort,
```
ddfreyne: I kept `rdoc` in there—I think it makes sense to have it in there.

ddfreyne:
RDoc stuff removed. To review!

bobthecow:
:+1: LGTM.

PR #592, opened 2015-05-17
Make globs and indentifiers with extensions the default

ddfreyne:
* [x] Make globs the default
* [x] Make identifiers with extensions the default

I don’t like the option names for enabling/disabling these (`pattern_syntax: glob/null` and `identifier_style: full/stripped`).

### DIFF
--- a/lib/nanoc/base/source_data/site.rb
+++ b/lib/nanoc/base/source_data/site.rb
@@ -24,7 +24,8 @@ module Nanoc::Int
       type: 'filesystem_unified',
       items_root: '/',
       layouts_root: '/',
-      config: {}
+      config: {},
+      identifier_style: 'full',
     }
 
     # The default configuration for a site. A site's configuration overrides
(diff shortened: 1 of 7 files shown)


bobthecow:
stripped is weird. null is weird. glob isn't the best, but at least it's accurate.

ddfreyne:
More ideas for identifiers:

* `identifier_type: with_ext`
* `identifier_type: without_ext`

And for patterns:

* `pattern_type: glob`
* `pattern_type: legacy_wildcard`
* `pattern_type: legacy`

(I prefer `_type` more than syntax/style.)

bobthecow:
I'm leaning toward:

 * `identifier_type: full`
 * `identifier_type: legacy`

and

 * `pattern_type: glob`
 * `pattern_type: legacy`

Let's make it very clear why the second option for each exists.

gpakosz:
I'm with @bobthecow

ddfreyne:
I’ll follow up with a second PR that changes the names to what @bobthecow suggests.

ddfreyne:
See #593.

PR #593, opened 2015-05-18
Rename identifier_style and pattern_syntax

ddfreyne:
Old:

* `identifier_style` &rarr; `"full"` / `"stripped"`
* `pattern_syntax` &rarr; `"glob"` / `nil`

New:

* `identifier_type` &rarr; `"full"` / `"legacy"`
* `string_pattern_type` &rarr; `"glob"` / `"legacy"`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,7 +7,7 @@ Changes:
 * Removed `ItemCollectionView#at`
 * Removed support for calling `ItemCollectionView#[]` with an integer
 * Renamed `identifier_style` to `identifier_type`, and made its values be `"full"` or `"legacy"`.
-* Renamed `pattern_syntax` to `pattern_type`, and made its values be `"glob"` or `"legacy"`.
+* Renamed `pattern_syntax` to `string_pattern_type`, and made its values be `"glob"` or `"legacy"`.
 
 Enhancements:
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,7 +6,8 @@ Changes:
 
 * Removed `ItemCollectionView#at`
 * Removed support for calling `ItemCollectionView#[]` with an integer
-* Renamed `identifier_type` to `identifier_type`, and made its values be `"full"` or `"legacy"`.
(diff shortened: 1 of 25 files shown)


NEWS.md
```diff
@@ -6,6 +6,7 @@ Changes:
 
 * Removed `ItemCollectionView#at`
 * Removed support for calling `ItemCollectionView#[]` with an integer
+* Renamed `identifier_type` to `identifier_type`, and made its values be `"full"` or `"legacy"`.
```
bobthecow: > Renamed `identifier_type` to `identifier_type`

:P

ddfreyne: Yup. I’m on to that right now. :)

ddfreyne:
Hm, two things:

* The constructor of Nanoc::Identifier has an optional `:style` argument. It currently defaults to `"legacy'`, but that’s not what we want. Maybe having it mandatory (as `:type`) makes sense, or we default to `"full"`.

* `pattern_type` might be better as `string_pattern_type`. Regexes are also patterns, but `pattern_type` doesn’t apply to that.

lib/nanoc/base/source_data/identifier.rb
```diff
@@ -16,10 +16,10 @@ def self.from(obj)
     end
 
     def initialize(string, params = {})
-      @style = params.fetch(:style, :stripped)
+      @style = params.fetch(:style, :legacy)
```
bobthecow: should this variable be renamed as well?

ddfreyne: Good catch.

bobthecow:
I'd say name the argument (and instance variable) `type` and default it to `full`.

ddfreyne:
Letting it default to `:full` seems like it belongs in a separate PR. I agree though.

Is this PR good to go?

ddfreyne:
Oh yeah, hm, I would like to replace `pattern_type` with `string_pattern_type`.

ddfreyne:
Renamed.

ddfreyne:
Merged. Hope it’s good :)

PR #599, opened 2015-05-25
Default filesystem data source

ddfreyne:
Three changes:

* renames `filesystem_unified` to `filesystem`, which is much nicer. I’ll keep `filesystem_unified` around.

* removes `filesystem_verbose`, which I believe nobody uses anymore. I’ve never seen it in the wild, and it is so annoying to work with that I doubt anyone deliberately still uses it. The default data source has been `filesystem_unified` for ~5 years. **Backwards incompatible.**

* removes obsolete code for a removed `--datasource` option in the `create-site` command.


bobthecow:
:+1:

PR #603, opened 2015-05-28
Require Ruby 2.2+

### LINKED ISSUE #597: Require Ruby 2.2+
Let’s have nanoc depend on Ruby 2.2 and up.

Ruby 2.0 will reach end of life in the next half year. JRuby 9000 is an implementation of Ruby 2.2.

ddfreyne:
This sets Ruby 2.2 as the minimum required version.

It also removes the magic encoding comments, which are no longer necessary because UTF-8 is now the default source encoding.

Fixes #597.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: ruby
 rvm:
   - 2.2.2
-  - 2.0.0
 branches:
   only:
     - master
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,3 @@
-# encoding: utf-8
-
 # Set up env
 $LOAD_PATH.unshift(File.expand_path(File.dirname(__FILE__) + '/lib'))
 require './test/gem_loader.rb'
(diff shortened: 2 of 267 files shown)


ddfreyne:
It turns out UTF-8 has been the default since Ruby 2.0—meaning even if we _don’t_ set the minimum version to 2.2, the magic comments can still go.

ddfreyne:
Things in Ruby 2.1 and 2.2 that I’d like to make use of:

* Optimized frozen strings (`"string"f`)
* Keyword arguments
* Refinements
* `Process.setproctitle` (just for fun)

TheLonelyGhost:
Also of note, Ruby 1.9.3 was EOL'ed in February, 2015. Definitely safe to take out UTF-8 encoding comments.

cdlm:
I still have an `encoding: UTF-8` line in `nanoc.yaml`. Presumably this relates to content and layouts… is it still needed?

ddfreyne:
@cdlm Yes. Ruby 2.0+’s default encoding for source files is UTF-8, but the external encoding isn’t. The `encoding: UTF-8` in the configuration file refers to the external encoding to use when reading files, not to the encoding of source files.

PR #613, opened 2015-06-01
Item rep writer

ddfreyne:
This is a refactoring that extracts item rep writing functionality into its own class. Quite WIP for the time being.

Philosophies in refactoring:

* Entities, such as item representations, should be data objects, and shouldn’t know how to interact with the outside world.
* Service classes, such as the item rep writer, can interact with the outside world (in this case, the filesystem).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ gem 'compass'
 gem 'coffee-script'
 gem 'coveralls', require: false
 gem 'erubis'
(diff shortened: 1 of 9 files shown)


ddfreyne:
This PR isn’t quite as extended as I would have liked, but it does improve the code quality a little bit and I will definitely continue to iterate on this.

PR #616, opened 2015-06-06
Improve checksums

ddfreyne:
* Specs have been migrated to RSpec and cleaned up.

* Checksums are generated from strings that are more readable (e.g. `String<foo>` rather than `Stringfoo`). This makes debugging easier. The final base 64 SHA1 digest is still the same size.

* There are custom checksums for `Symbol`, `Time`, `Numeric`, `nil`, `true` and `false`, so we don’t have to rely on marshalling anymore.

* There are custom checksums for all views (`Nanoc::ItemView`, `Nanoc::LayoutView`, `Nanoc::ConfigView`), as well as for `Nanoc::Int::IdentifiableCollection`. This is expected to reduce cases where checksums incorrectly differ.

* Item and layout checksums now take the identifier into account. (I’m not sure why this wasn’t the case before.)

* `#marshal_dump` and `#marshal_load` implementations on Item and Layout have been removed, because they are no longer used.


bobthecow:
:+1:

PR #617, opened 2015-06-06
Remove Item/Layout attribute shorthands

ddfreyne:
* `Item` and `Layout` no longer have `#[]` and `#[]=`. They do not serve a purpose anymore; the shorthand is handled by the views.

* Dependency tracking (“visit started” and “visit ended” notifications) should happen in the views. This PR makes that change.

* Strangely, no dependency tracking was done for attributes in layouts. This PR also fixes that issue.

* As expected, some helper/filter tests use the entities rather than the views, causing their tests to break after this change. The tests have been updated to use views rather than entities.

To do (in a separate PR):

* Move common `ItemView` and `LayoutView` method into module or superclass (`#[]`, `#fetch`, `#key?`). Same goes for the mutable variants (`#[]=`, `#update_attributes`).

### DIFF
--- a/lib/nanoc/base/views/item.rb
+++ b/lib/nanoc/base/views/item.rb
@@ -31,10 +31,11 @@ module Nanoc
 
     # @see Hash#fetch
     def fetch(key, fallback = NONE, &_block)
-      res = @item[key] # necessary for dependency tracking
+      Nanoc::Int::NotificationCenter.post(:visit_started, @item)
+      Nanoc::Int::NotificationCenter.post(:visit_ended,   @item)
 
       if @item.attributes.key?(key)
-        res
+        @item.attributes[key]
       else
         if !fallback.equal?(NONE)
           fallback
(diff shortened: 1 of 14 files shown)


lib/nanoc/base/views/item.rb
```diff
@@ -31,10 +31,11 @@ def identifier
 
     # @see Hash#fetch
     def fetch(key, fallback = NONE, &_block)
-      res = @item[key] # necessary for dependency tracking
+      Nanoc::Int::NotificationCenter.post(:visit_started, @item)
+      Nanoc::Int::NotificationCenter.post(:visit_ended,   @item)
```
bobthecow: Why is the `:visit_ended` before it has actually ended?

bobthecow:
:+1:

lib/nanoc/base/views/item.rb
ddfreyne: The “visit” ends before the method returns. There’s nothing to be done _during_ the visit in this case.

There’s really only one place where stuff needs to be tracked during a visit (between start/end) and that is when compiling an item.

Having a `:visited` shorthand could be useful, I guess!

bobthecow: Yeah, that's kind of what I was wondering.

PR #618, opened 2015-06-06
Document class

ddfreyne:
* Create Document supertype for `Item` and `Layout`. These two classes are quite similar (have content, attributes, and identifier), so making them of a similar type makes sense.

* Create DocumentView supertype. This DRYs up the implementation of `ItemView` and `LayoutView`, and makes `#fetch` and `#key?` available in `LayoutView`.


bobthecow:
:+1:

PR #621, opened 2015-06-06
Content entity

ddfreyne:
This PR extracts the logic for dealing with textual vs binary content into a proper content entity.

Backwards compatible, too!

All kinds of content, including textual content, have a filename, containing the absolute path to the file that contains the content. This is necessary in order for some filters to work: for example, the Sass filter uses it to find items.

This also cleans some related stuff:

* Legacy `mtime` handling (it’s just an attribute, and it should be treated as such).

* `raw_filename` is now part of the content, rather than being part of the item.

* Layouts can now theoretically be binary. I don’t think this is particularly useful, but there’s no reason why they shouldn’t be able to. (The filesystem data source still prevents binary layouts from being generated though.)

Future work (out of scope for this PR):

* Use content entities for representations/snapshots instead of only items and layouts.

* Enforce absolute filenames.

* (Maybe) extend the kinds of content to more than just textual and binary content at some point. For example, parsed XML DOM, or Kramdown internal representation. This seems useful, although probably with only marginal benefits.

Documentation changes:

* `#new_item` argument `:filename`
* `#create` argument `:filename`


ddfreyne:
My attempt at not refactoring too much before the final 4.0 release isn’t really working out well, but I hope it’s for the better :)

PR #622, opened 2015-06-07
Require absolute filenames

ddfreyne:
This removes a dependency on the current working directory.

Documentation changes:

* Filenames used when generating items or layouts need to be absolute.

### DIFF
--- a/lib/nanoc/base/entities/content.rb
+++ b/lib/nanoc/base/entities/content.rb
@@ -15,6 +15,10 @@ module Nanoc
 
       # @param [String, nil] filename
       def initialize(filename)
+        if filename && !filename.start_with?('/')
+          raise ArgumentError, 'Content filename is not absolute'
+        end
+
         @filename = filename
       end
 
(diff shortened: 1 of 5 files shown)


lib/nanoc/base/entities/content.rb
```diff
@@ -15,6 +15,10 @@ class Content
 
       # @param [String, nil] filename
       def initialize(filename)
+        if filename && !filename.start_with?('/')
+          raise ArgumentError, 'Content filename is not absolute'
+        end
```
bobthecow: does this work on the windoze?

ddfreyne: Probably not. :(

ddfreyne: And I can’t verify/test, because I don’t have access to a Windows machine.

Crazycatz00: I can verify this does not work on windows; it throws for the filename "D:/site/content/htaccess.noext".
I was however able to get this to work using `!Pathname.new(filename).absolute?` instead of `!filename.start_with?('/')`.

gpakosz:
I find it indeed better to use `Pathname`

lib/nanoc/base/entities/content.rb
ddfreyne: Fix in #656.

PR #624, opened 2015-06-07
Fix freezing

ddfreyne:
The freeze tests didn’t check whether an error was raised at all; just that *if* an error is raised, it is the correct one.


bobthecow:
:+1:

ddfreyne:
Custom RSpec matchers are :heart: though.

PR #638, opened 2015-06-14
Full site loader, attempt 2

ddfreyne:
See #636, #637.


ddfreyne:
It turns out that the problem I saw with the original (reverted) PR was some local changes made to the nanoc.ws site—not with the code in nanoc.

Taking the time to write better tests in this PR anyway!

PR #641, opened 2015-06-16
Fix RuboCop and Coveralls integration

ddfreyne:
The Rubocop integration was failing because because of the call to `#exit`. SimpleCov/Coveralls should now also work properly.

Additionally, this PR simplifies the Rakefiles. It also removes `LOAD_PATH` magic.

### DIFF
--- a/lib/nanoc/cli/cleaning_stream.rb
+++ b/lib/nanoc/cli/cleaning_stream.rb
@@ -132,6 +132,11 @@ module Nanoc::CLI
       @stream.external_encoding
     end
 
+    # @see ARGF.set_encoding
+    def set_encoding(*args)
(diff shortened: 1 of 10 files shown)


lib/nanoc/cli/cleaning_stream.rb
```diff
@@ -132,6 +132,11 @@ def external_encoding
       @stream.external_encoding
     end
 
+    # @see ARGF.set_encoding
+    def set_encoding(*args)
```
ddfreyne: Necessary because Rake calls this for some reason.

denis-soundcloud:
That quite clearly did not work; see [Coveralls page for nanoc](https://coveralls.io/r/nanoc/nanoc). Will revisit.

PR #642, opened 2015-06-17
Check for Gemfile

ddfreyne:
This shows the following message when nanoc is invoked without `bundle exec`, and a _Gemfile_ is present:

> A Gemfile was detected, but Bundler is not loaded. This is probably not what you want. To run nanoc with Bundler, use `bundle exec nanoc`.


ddfreyne:
Not sure how to test this (is this even testable?)

PR #646, opened 2015-06-18
Checksummer with refinements

ddfreyne:
Experiment.


ddfreyne:
FYI, CodeClimate actually says the new code is worse.

ddfreyne:
Experiment results: refinements work, although I don’t know how performant they are, and I don’t think it improves over the original code.

ddfreyne:
Doesn’t provide any real benefit; closing.

PR #650, opened 2015-06-20
Fix sidebar links in default site

ddfreyne:
* Added a link to the about page.

* Removed the link to the wiki. The wiki is not particularly useful, and linking to it makes it feel like it is an important part of the documentation or community.

* Pointed all links to `v4.nanoc.ws` rather than `nanoc.ws`. This is necessary because the links otherwise point to old documentation. This is temporary and needs to be changed before the v4 final release.


bobthecow:
:+1:

PR #653, opened 2015-06-21
Refactor compiler (part 1 of…)

ddfreyne:
This started as a refactoring of the compiler loading mechanism, but turned more into a refactoring of the rule loading logic.

The `RulesCollection` class is now an entity with no external dependencies.

There’s also some smaller minor changes to the compiler. For instance, there is now an `ItemRepSelector` (yields item reps to compile based on dependencies) and an `ItemRepRouter` (assigns paths to item reps).


bobthecow:
:+1: lgtm

PR #655, opened 2015-06-21
Create item rep repo

ddfreyne:
This is an attempt at removing the item reps collection from `ItemRep`. This should make it easier to move to streaming data sources (hopefully).


ddfreyne:
Replaced by #668.

PR #658, opened 2015-06-27
Use keyword arguments

ddfreyne:
Ruby 2.2 supports keyword arguments, which has two advantages:

* It makes existing code more compact. No more `params.fetch` necessary.
* It increases reliability because unrecognised arguments will cause an `ArgumentError`.

Some occurrences of `params` cannot be replaced with keyword arguments—filters, for instance.

This change also removes a deprecated extra unused `params` argument in the `CodeSnippet` constructor.

To do:

* [x] Use keyword arguments everywhere
* [x] Update documentation from `@option` to `@param`

### DIFF
--- a/lib/nanoc/base/compilation/compiler.rb
+++ b/lib/nanoc/base/compilation/compiler.rb
@@ -72,14 +72,14 @@ module Nanoc::Int
 
     # @group Public instance methods
 
-    def initialize(site, rules_collection, params = {})
+    def initialize(site, rules_collection, compiled_content_cache:, checksum_store:, rule_memory_store:, rule_memory_calculator:)
       @site = site
       @rules_collection = rules_collection
 
-      @compiled_content_cache = params.fetch(:compiled_content_cache)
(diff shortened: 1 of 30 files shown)


ddfreyne:
Build on Travis CI fails because Less segfaults. Will fix later.

lib/nanoc/base/compilation/compiler.rb
```diff
@@ -72,14 +72,14 @@ class Compiler
 
     # @group Public instance methods
 
-    def initialize(site, rules_collection, params = {})
+    def initialize(site, rules_collection, compiled_content_cache:, checksum_store:, rule_memory_store:, rule_memory_calculator:)
```
bobthecow: Oh, I didn't realize keyword args could be empty. I've always defaulted them to something.

ddfreyne: Yup, new in Ruby 2.1.

PR #671, opened 2015-07-08
fixed typo in class definition of shell command module

jimjimovich:
Running nanoc shell was producing the following error:
NoMethodError: undefined method `env_for_site' for Nanoc::CLI::Commands::Shell:Class

I changed the method env_for to env_for_site and now everything works fine.


ddfreyne:
Thanks for the fix!

The way the CLI is tested right now is unfortunately a little brittle. I’ll look into improving this, so that every command is tested at least on a high level by ensuring that invoking the command works.

ddfreyne:
I’ve opened #672 which contains this fix + a high-level test.

Issue #674, opened 2015-07-10
Identifier.split used to work, now it doesn't


pitosalas:
This might not be a bug, just an upgrade tip. But if I try to do a split on an identifier I get:

"NoMethodError: undefined method `split' for <Nanoc::Identifier type=legacy "/bootstrap/css/custom/">:Nanoc::Identifier"

ddfreyne:
A fix is to replace `identifier.split` with `identifier.to_s.split`, but considering this is fairly common use case, I believe `Identifier` either needs to support `#split` (just like `String#split`), or gain a `#components` method, which would do the splitting by slash.

See also [this thread](https://groups.google.com/d/msg/nanoc/yz2Czn6UEr0/nKsRhqKipTgJ).

ddfreyne:
See #677 for an implementation of `#components`. Does this fit your use case?

I’m more in favour of not adding `#split` to `Identifier`, and rather go with the `identifier.to_s.split` approach. (Until `#components` is available, that is.)

ddfreyne:
I’ll close this issue, because I’d prefer not to have `#split` on `Identifier`, and the upgrade guide’s [troubleshooting section](http://v4.nanoc.ws/doc/nanoc-4-upgrade-guide/#troubleshooting) explains this already.

pitosalas:
I agree... Once I figured it out, no big deal!

-- Pito
p.s. sent from my iPhone, so excuse brevity, typos, etc.

PR #677, opened 2015-07-11
Add Identifier#components

ddfreyne:
This provides a slightly nicer-to-use alternative to `identifier.to_s.split`.

The return value is such that `'/' + identifier.components.join('/')` is identical to `identifier.to_s`.

See #674.


ddfreyne:
Still unsure whether the components of `/` should be an empty list, or a list with an empty string.

Issue #664, opened 2015-06-30
Better consistensy between Filesystem and Nanoc::Identifier (Nanoc 4.x)


barraq:
`Nanoc::Filesystem` currently provides one of the following options: `allow_periods_in_identifiers`. As said in the documentation:

    # Whether to allow periods in identifiers. When turned off, everything
    # past the first period is considered to be the extension, and when
    # turned on, only the characters past the last period are considered to
    # be the extension. For example,  a file named “content/about.html.erb”
    # will have the identifier “/about/” when turned off, but when turned on
    # it will become “/about.html/” instead.
    allow_periods_in_identifiers: false

When using this option with the `full` identifier the behavior of Nanoc::Identifier is to my point of view misleading: let me explain. Let `toto.html.haml` be a file in your filesystem.
Suppose that you have the following rule:

```ruby
route '/toto.*' do
  puts item[:extension]
  puts item.identifier
  puts item.identifier.with_ext('html')
  '/toto.html'
end
```

The following is outputted

```shell
html.haml
/toto.html.haml
/toto.html.html
```

Therefore

- item[:extension] returns `html.haml` which is what is expected
- item.identifier returns `/toto.html.haml` which is what is expected in `full` mode
- however `item.identifier.with_ext('html')` returns `toto.html.html` which is not what I expect when using the `allow_periods_in_identifiers=false` since  item[:extension] returns `html.haml`. It would be better if it returned `toto.html` therefore using:

    ```ruby
    # Returns a regex that is used for determining the extension of a file
    # name. The first match group will be the entire extension, including the
    # leading period.
    def extension_regex
      if @config && @config[:allow_periods_in_identifiers]
        /(\.[^\/\.]+$)/
      else
        /(\.[^\/]+$)/
      end
    end
    ```
Maybe the option `allow_periods_in_identifiers` lose some meaning when in `full` identifier mode but then it would be really cool to provide another option so that we can use:  `item.identifier.with_ext('html')` in a way that is coherent with what item[:extension] returns.

What do you think ?

Best

ddfreyne:
Yeah, definitely something fishy going on here… I think `allow_periods_in_identifiers` should be considered to be `false` when for full identifiers.

ddfreyne:
I’ll think about this a bit more though.

barraq:
@ddfreyne should not the Nanoc::Identifier module (for an Item) be relying on the Nanoc::FileSystem module (to which this Item belong) to manipulate extension ? 

[...]

PR #679, opened 2015-07-17
addresses #664

barraq:
This is a first draft to introduce:

- `#exts` method (returning a list of extensions)
- `#without_exts` (removes all extensions)

methods to Nanoc::Identifier

### DIFF
--- a/lib/nanoc/base/entities/identifier.rb
+++ b/lib/nanoc/base/entities/identifier.rb
@@ -147,6 +147,26 @@ module Nanoc
       s && s[1..-1]
     end
 
+    # @return [String]
+    def without_exts
+      extname = exts.join('.')
+      if extname.size > 0
+        @string[0..-extname.size - 2]
+      else
+        @string
+      end
+    end
+
+    # @return [Array] List of extensions, without a leading dot.
+    def exts
+      unless full?
+        raise UnsupportedLegacyOperationError
+      end
+
+      s = File.basename(@string)
+      s ? s.split('.', -1).drop(1) : []
+    end
+
     def to_s
       @string
     end
--- a/spec/nanoc/base/entities/identifier_spec.rb
+++ b/spec/nanoc/base/entities/identifier_spec.rb
@@ -375,4 +375,70 @@ describe Nanoc::Identifier do
       it { is_expected.to eql('md') }
     end
   end
+
+  describe '#without_exts' do
+    subject { identifier.without_exts }
+
+    context 'legacy type' do
+      let(:identifier) { described_class.new('/foo/', type: :legacy) }
+
+      it 'raises an error' do
+        expect { subject }.to raise_error(Nanoc::Identifier::UnsupportedLegacyOperationError)
+      end
+    end
+
+    context 'identifier with no extension' do
+      let(:identifier) { described_class.new('/foo') }
+
+      it 'does nothing' do
+        expect(subject).to eql('/foo')
+      end
+    end
+
+    context 'identifier with one extension' do
+      let(:identifier) { described_class.new('/foo.md') }
+
+      it 'removes the extension' do
+        expect(subject).to eql('/foo')
+      end
+    end
+
+    context 'identifier with multiple extensions' do
+      let(:identifier) { described_class.new('/foo.html.md') }
+
+      it 'removes the extension' do
+        expect(subject).to eql('/foo')
+      end
+    end
+  end
+
+  describe '#exts' do
+    subject { identifier.exts }
+
+    context 'legacy type' do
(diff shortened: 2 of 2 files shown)


barraq:
@ddfreyne actually I wonder if `#without_ext` and `#without_exts` should not be returning `Nanoc::Identifier.new @string` instead of `@string`.
Actually an identifier without ext(s) should still be an identifier right ?

ddfreyne:
@barraq Since an identifier is used to identify items/layouts, and these methods are used for creating an output path to which the item should be written to, I don’t feel that these methods should return an identifier.

They could potentially return a “path” entity, but I feel that a string is more appropriate.

(I don’t like `Pathname` for this purpose, because it has methods that are really not appropriate in a Nanoc context, such as `#rmdir` and `#rename`, and I feel that the API is generally clunky. `#extname` is a strange method name, and inconsistent with `#sub_ext`.)

I also feel that the behavior of `#with_ext` is a little odd. Perhaps it needs to be replaced with `#replace_ext` (replaces a single extension), `#replace_exts` (replaces all extensions), `#remove_exts` (removes all extensions), and `#add_ext` (which appends an extension).

I do feel it’s a little late in the Nanoc 4 design process (release candidate!) to change this behavior still. But then again, I’d rather have this discussion *now* rather than after the final release.

barraq:
> Since an identifier is used to identify items/layouts, and these methods are used for creating an output path to which the item should be written to, I don’t feel that these methods should return an identifier.

@ddfreyne I was asking that question because I was wondering if my PR was correct.
Let me explain. Actually if you want to be able to chain `#without_exts.with_ext`, as you proposed below: 

```ruby
Nanoc::Identifier.new('/toto.html.haml').without_exts.with_ext('html')
# => '/toto.html'

Nanoc::Identifier.new('/toto.html.haml').without_ext.with_ext('html')
# => '/toto.html'

Nanoc::Identifier.new('/toto.html.haml').with_ext('html')
# => '/toto.html.html'
```

Then you have two solution I guess:

- either `#without_exts` and "#without_ext" return a Nanoc::Identifier, which, to me, is more logic (for the layout you could then do ```layout.identifier.without_ext.to_s``` which is nice to me)
- or you have to extend `String` but I really don't like it...
- or you think of something else ?

> I also feel that the behavior of #with_ext is a little odd. Perhaps it needs to be replaced with #replace_ext (replaces a single extension), #replace_exts (replaces all extensions), #remove_exts (removes all extensions), and #add_ext (which appends an extension).

Totally agree ! You can still keep `#with_ext` for backward compatibility (and warn users that it is deprecated). `#with_ext` will just be `#remove_ext.add_ext` or `#replace_ext`.

> I do feel it’s a little late in the Nanoc 4 design process (release candidate!) to change this behavior still. But then again, I’d rather have this discussion now rather than after the final release.

I understand... do you have an idea on how much people already switched to Nanoc::4 ? I guess that if we do these changes now it will not have such a negative impact... Also these changes are not braking anything if we keep `#without_ext` and `#without_exts` for compatibility purpose, right ?

ddfreyne:
Sorry for getting back to you with such a delay! I got held up by other things.

> Actually if you want to be able to chain `#without_exts.with_ext`, …

Perhaps it makes no sense to allow this, and go with a simpler approach where we *only* have `#without_ext` and `#without_exts`. Adding an extension could be done using `#+`, like this:

```ruby
Nanoc::Identifier.new('/toto.html.haml').without_exts + '.html'
# => '/toto.html'

Nanoc::Identifier.new('/toto.html.haml').without_exts + '/index.html'
# => '/toto/index.html'
```

> You can still keep `#with_ext` for backward compatibility (and warn users that it is deprecated)

I’d prefer to not introduce deprecations in 4.0.0 final, but there could be an appropriate error message when using `#with_ext` in Nanoc 4 release candidate 3.

>> I do feel it’s a little late in the Nanoc 4 design process (release candidate!) to change this behavior still. But then again, I’d rather have this discussion now rather than after the final release.

> I understand... do you have an idea on how much people already switched to Nanoc::4 ? I guess that if we do these changes now it will not have such a negative impact... Also these changes are not braking anything if we keep `#without_ext` and `#without_exts` for compatibility purpose, right ?

I’ll probably make this breaking change anyway, because it seems worth it to me.

ddfreyne:
FYI, this PR gets a :+1:—the discussion about what to do in addition to this work.

ddfreyne:
Ah—could you make this PR on `release-4.0.x` instead? This way, it’ll end up in 4.0 rather than 4.1.

barraq:
Thank you @ddfreyne ! I'll do the PR on `release-4.0.x` then.

PR #687, opened 2015-08-04
Make base_url optional in #tags_for

croath:
While using version 4.0.2rc2, one error occurred after `Tagging` helper and `tags_for` method were added into my code, which told me a KeyError around key `base_url`.

I checked the source in release-4.0.x and found that `base_url` must be provided or there will be an error here. It's never been listed in the document or the official tutorial.

In my mind it's not so important that users MUST have a link to a tag, at least there should be a default tag link url.

Thank you so much for your review.

### DIFF
--- a/lib/nanoc/helpers/tagging.rb
+++ b/lib/nanoc/helpers/tagging.rb
@@ -16,7 +16,9 @@ module Nanoc::Helpers
     # HTML-escaping rules for {#link_for_tag} apply here as well.
     #
     # @option params [String] base_url The URL to which the tag will be appended
-    #   to construct the link URL. This URL must have a trailing slash.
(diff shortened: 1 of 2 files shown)


ddfreyne:
I think it would make more sense to let `base_url` be optional, and not generate a link if a `base_url` is not provided.

croath:
@ddfreyne Yes you are right... I mean, it really make sense. Check the appending commit :smile:

ddfreyne:
This looks good! Could you make sure to update the tests too?

croath:
Test function added!

ddfreyne:
Looks good. Thanks!

croath:
With pleasure :smile:

PR #706, opened 2015-10-19
Add `#each` to `ConfigView`

### LINKED ISSUE #705: Add `each` to ConfigView?
I was toying around with jumping from Nanoc3 to 4 and found that one of my preprocessors has stopped working. It depends on iterating over the existing config keys to set some other information up.

Would there be any interest in a PR that adds `each` to `ConfigView`?

gjtorikian:
Closes #705 

I actually wrote a little dinky test for this in *test_compiler_dsl.rb*: 

``` ruby
  def test_config_iteration
    hash = { :key_one => 1, :key_two => 2}
    @config = Nanoc::ConfigView.new(hash, nil)
    @config.each do |k, v|
      puts k, v
    end
  end
```

It works, but because no other `ConfigView` syntaxes are tested (and the fact that STDOUT is muted) I decided not to include it. I could if there's a better way to work this.

### DIFF
--- a/lib/nanoc/base/views/config_view.rb
+++ b/lib/nanoc/base/views/config_view.rb
@@ -36,5 +36,10 @@ module Nanoc
     def [](key)
       @config[key]
     end
+
+    # @see Hash#each
+    def each(&block)
+      @config.each(&block)
+    end
   end
 end


ddfreyne:
Oh, the lack of unit tests for `ConfigView` is an oversight on my part. I’ll fix in a future PR, but for now, I’m merging this!

ddfreyne:
So apparently there *are* `ConfigView` tests—except that they’re confusingly in `spec/nanoc/base/views/config_spec.rb` rather than <code>…/config_<b>view</b>_spec.rb</code>. So I renamed the files (f0cdb83) and added a test (aacdc00d9423cf487cb57f7e249c13b3ce967a86).

PR #710, opened 2015-10-21
Let #parent/#children raise for non-legacy identifiers

### LINKED ISSUE #709: #parent and #children should raise for full identifiers
When attempting to get the parent or children (using `#parent` resp. `#children`) of an item/layout that has a full identifier (e.g. `/foo.md` rather than the legacy `/foo/`), an exception should be thrown. These two methods do not yield useful results for non-legacy identifiers.

ddfreyne:
Fixes #709.

It also adds:

* a test for `#children` (which was still pending)
* `Identifier#legacy?` because it’s useful
* tests for `#full?` and `#legacy?`

### DIFF
--- a/lib/nanoc/base/entities/identifier.rb
+++ b/lib/nanoc/base/entities/identifier.rb
@@ -87,6 +87,12 @@ module Nanoc
       @type == :full
     end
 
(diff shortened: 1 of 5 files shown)


ddfreyne:
CC @nanoc/contributors

PR #712, opened 2015-10-22
Only attempt to clean if the item is a string

gjtorikian:
I've noticed that if I try to do some quick debugging—like, `puts @config`, this little helper will complain about a missing `gsub`, because `s` is a Hash or a String or something.

I think it's necessary every now and then to not assume everything being printed is a string, so this PR protects against that.


ddfreyne:
Makes sense, thanks!

Issue #704, opened 2015-10-09
Multiple "content_for" calls should concatenate


Fjan:
Currently, when you call `content_for` twice on a page the second call will overwrite the content captured for the first one. It would be more useful if they concatenated, just like the behaviour of content_for in Rails. Consider this example that would be made possible:
```
<% content_for :head do %>
   <link ref="...">
<% end %>
```
And somewhere else you could do:
```
<% content_for :head do %>
   <script src="...">
<% end %>
```
And then put a single `<%= content_for :head %>` in the layout.

It would actually be a very simple change in nanoc/helpers/capturing.rb. Just replace line 45 with this:
```
if @store[item.identifier][name] then
  @store[item.identifier][name] << content
else 
  @store[item.identifier][name] = content
end
```

ddfreyne:
This wouldn’t work out of the box without breaking backwards compatibility, but it could work with a param like `:append`:

```erb
<% content_for :head, append: true do %>
   <script src="...">
<% end %>
```

vs

```erb
<% content_for :head, append: false do %>
   <script src="...">
<% end %>
```

vs

```erb
<% content_for :head do %>
   <script src="...">
<% end %>
```

(default: replace rather than append)

Fjan:
Do you think anybody uses the current behaviour of calling content_for twice with the same symbol, which would erase the contents of first call? I find it hard to imagine a use case for that. I think most people simply call it once so the issue never arises, so a warning of the changed behaviour in the upgrade notes would suffice. Of course, if I'm overlooking a way in which the current behaviour might be useful, then adding it as an option would be better.

ddfreyne:
It’s not possible to change the current behavior without breaking backwards compatibility, and since Nanoc 4 is so close to a final release, this behavior is not going to change before Nanoc 5.0.

ddfreyne:
FYI, I’d certainly welcome a PR that adds the new functionality (in a backwards-compatible way); it’ll end up in Nanoc 4.1 if you do so!

Fjan:
Ok, I don't mind making a PR, but I really don't like adding a parameter for an option for which no one can think of a use case. I'm fine with using the 1-line monkey patch until the behaviour can change in 5.0

ddfreyne:
I’m OK with a change in behavior in 5.0!

gpakosz:
Can't you add an `:append` parameter that's opt-in? Would keep the default behavior untouched while enabling the new use case

Fjan:
[...]

PR #713, opened 2015-10-30
Make content_for raise an exception in case of silently overwriting content

Fjan:
This addresses #704

### DIFF
--- a/lib/nanoc/helpers/capturing.rb
+++ b/lib/nanoc/helpers/capturing.rb
@@ -33,7 +33,6 @@ module Nanoc::Helpers
 
       def []=(item, name, content)
         @store[item.identifier] ||= {}
-        # @store[item.identifier][name] = content
         if @store[item.identifier][name]
           return if @store[item.identifier][name] == content
           raise KeyError, "Content_for was called twice with the same key: #{name}, this would overwrite content for the first call"
--- a/lib/nanoc/helpers/capturing.rb
+++ b/lib/nanoc/helpers/capturing.rb
@@ -33,7 +33,14 @@ module Nanoc::Helpers
 
       def []=(item, name, content)
         @store[item.identifier] ||= {}
-        @store[item.identifier][name] = content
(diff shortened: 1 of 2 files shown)


ddfreyne:
Could you add tests for this, and remove the commented-out code?

Fjan:
OK, done

test/helpers/test_capturing.rb
```diff
@@ -101,6 +101,32 @@ def test_different_sites
...
   end
 
+  def test_content_for_duplicate_symbols
+    with_site do |_site|
+      # Prepare
+      File.open('lib/helpers.rb', 'w') do |io|
+        io.write 'include Nanoc::Helpers::Capturing'
+      end
+      File.open('content/includer.erb', 'w') do |io|
+        io.write '[<%= content_for(@items.find { |i| i.identifier == \'/includee/\' }, :blah) %>]'
```
ddfreyne: Replace

```ruby
@items.find { |i| i.identifier == \'/includee/\' }
```

with

```ruby
@items['/includee/']
end

File: lib/nanoc/helpers/capturing.rb [hunk 2]
```diff
@@ -33,7 +33,12 @@ def initialize
 
       def []=(item, name, content)
         @store[item.identifier] ||= {}
-        @store[item.identifier][name] = content
+        if @store[item.identifier][name]
+          return if @store[item.identifier][name] == content
+          raise KeyError, "Content_for was called twice with the same key: #{name}, this would overwrite content for the first call"
```
ddfreyne: I don’t think `KeyError` is correct—that is an exception used when the value of a given key isn’t found. I think raising a string is OK.

test/helpers/test_capturing.rb
Fjan: Ok, I changed that, but note that all other tests in that file use the other style

ddfreyne:
Thanks!

PR #717, opened 2015-11-04
Add DateTime handling to attribute_to_time

rosetree:
This enhances `attribute_to_time` which previously removed all time
information from a DateTime. This happened because DateTime identifies
itself as a Date.

    2.2.3 :001 > require 'date'
     => true
    2.2.3 :002 > DateTime.now.is_a?(Date)
     => true

Up to now, if `time.is_a?(Date)` is true, `attribute_to_time` will
create a new Time, with just year, month and day set. To fix that
behaviour, this commit adds a new condition at the beginning of
`attribute_to_time`, which uses `time.to_time` as long as `time` is a
DateTime.

### DIFF
--- a/lib/nanoc/helpers/blogging.rb
+++ b/lib/nanoc/helpers/blogging.rb
@@ -377,14 +377,15 @@ module Nanoc::Helpers
       'tag:' + hostname + ',' + formatted_date + ':' + base_dir + (item.path || item.identifier.to_s)
     end
 
-    # Converts the given attribute (which can be a string, a Time or a Date)
-    # into a Time.
+    # Converts the given attribute (which can be a string, a Time, a Date or a
+    # DateTime) into a Time.
     #
-    # @param [String, Time, Date] time Something that contains time
+    # @param [String, Time, Date, DateTime] time Something that contains time
     #   information but is not necessarily a Time instance yet
     #
     # @return [Time] The Time instance corresponding to the given input
     def attribute_to_time(time)
+      time = time.to_time if time.is_a?(DateTime)
       time = Time.local(time.year, time.month, time.day) if time.is_a?(Date)
       time = Time.parse(time) if time.is_a?(String)
       time


ddfreyne:
Looks good!

I realised that there’s no tests for `#attribute_to_time`—not sure how that happened, but presumably it’s because this function is tested implicitly. If you want to, you can create a spec for this function. If not, feel free to say no—I’ll take it up soon!

rosetree:
I’d love to add tests for `#attribute_to_time`, but actually I’m not really sure how to do this. That shouldn’t be a huge problem, though. I’m going to learn and do this on Saturday, if that’s not to late. Maybe I’d come with some questions back to you, @ddfreyne. Should this be tested with values of example items or with static objects? Could you point me to a similar test within Nanoc, that I could use as a reference? Thanks :smiley_cat:.

ddfreyne:
I went ahead and added a spec myself—hope you don’t mind! You can find the commit at https://github.com/nanoc/nanoc/commit/df20f2eb2b908d0647f141dd39e8594f90c91e8f.

Some notes (for learning):

* `mod` is an instance of a class that includes `Nanoc::Helpers::Blogging`. It’s a little unusual, because testing a module directly isn’t that simple.

* `subject` defines what needs to be tested: in this case, `mod.attribute_to_time(arg)`, where `arg` is the argument which takes on different values in the different `context`s.

* `context` defines different context—called with `Time`, `Date`, `DateTime` and `String`. If the method under test would’ve accepted an integer, with a Unix timestamp, there’d also be a context testing an int.

* `let` gives `arg` a value.

* `it { is_expected.to eql(something) }` describes what the subject is expected to be. This is where the actual checks are. This is a shorthand form for the following:

  ```ruby
  it "returns a correct Time instance" do
    expect(subject).to eql(something)
  end
  ```

I quite like RSpec. The [Better Specs](http://betterspecs.org/) site is very helpful for making the best use of RSpec. Nanoc 3.x used Minitest, but I’m migrating parts to RSpec, and new specs are all written using RSpec.

rosetree:
I don’t mind at all. Thank you for providing this explanation. :)

ddfreyne:
There’s a follow-up PR here: #724 — turns out the original code did not properly tackle timezones, and Travis CI builds failed.

PR #719, opened 2015-11-07
Fix indentation of default config

remko:


ddfreyne:
Could you make this PR on `release-4.0.x` instead? This way, it’ll end up in 4.0.1.

remko:
AFAICT, this was introduced with a feature that's not on the release branch yet.

ddfreyne:
Ah, correct. :)

Issue #711, opened 2015-10-21
Can I implement postprocessing?


gjtorikian:
I'm curious to know if that'd be a useful addition. Which is a humble way of saying, "I need it, and I think others might need it too!" :grinning: 

Essentially a postprocess would occur *after* all compilation and purging. It would rerun when/if items changed. I'd probably take a peek at the implementation for preprocess and crib heavily from it.

ddfreyne:
Having a post-process block has been on the wishlist for a while. The wiki page containing the archived Trello board even has a [Postprocess section](https://github.com/nanoc/nanoc/wiki/%5BARCHIVED%5D-nanoc-4-Trello-board#postprocess-block).

What’s your use case specifically?

A postprocess block could be as simple as a block that’s executed at the end. For instance:

```ruby
postprocess do
  @items.each do |item|
    $search_index.update(
      item.path,
      item.compiled_content(snapshot: :last),
    )
  end
end
```

This could be extended by letting item instances (and item rep instances) within the postprocess block have a `#updated?` and/or `#created?` methods, so you can act on updated/created items only. For instance:

```ruby
postprocess do
  @items.select(&:updated?).each do |item|
    $search_index.update(
      item.path,
      item.compiled_content(snapshot: :last),
    )
  end
end
```

gjtorikian:
My specific use-case is in generating additional content that falls out of "normal" compilation. For example, given a Markdown file that looks like this:

``` markdown
title: A title
redirects_from:
  - /some-title/
  - /nowhere/
```

I want two new HTML files in *output/some-title/index.html* and *output/nowhere/* with some client-side set up, like:

```
<meta http-equiv=refresh content="0; url=/a-title/}">
<h1>Redirecting...</h1>
<a href="/a-title/">Click here if you are not redirected.</a>
<script>location='/a-title/'</script>
```

I started doing this in a preprocess, and was adding new `Nanoc::Int::Item`s to the mutable collection, but then the build started complaining and wanting to track those new items with compile and routing steps. My workaround for that started to feel hokey and I just wanted to be able to operate with the same data, but outside all the Rules.


> This could be extended by letting item instances (and item rep instances) within the postprocess block have #updated? and/or #created? methods, so you can act on updated/created items only.

I like that a lot.

bobthecow:
You can route and compile the original item two times using a rep. It works rather well for things like this.

gjtorikian:
[...]

Issue #571, opened 2015-05-11
View API too broad for preprocessing


ddfreyne:
The `MutableItemView` class has methods such as `#path` and `#compiled_content`, which only make sense in a context where the item reps are available.

Perhaps a `MutableItemPreprocessorView` is in order.

ddfreyne:
Fixed in #768.

PR #726, opened 2015-11-12
Implement postprocessing

gjtorikian:
This is a start to the implementation of a `postprocessor` block. I'm opening it up somewhat early to get some feedback on whether I'm on the right track.

I mostly copied the work going on for the `preprocessor` block. I'd like to have an immutable items collection, with the addition of methods like `updated?` and `added?`. That's what *attributed_item_collection_view.rb* is for.

Closes https://github.com/nanoc/nanoc/issues/711.


ddfreyne:
Looks good so far!

I think there’s potential for quite a bit of refactoring here (e.g. unify code for pre- and postprocessor, move tests into specs) but that can wait.

gjtorikian:
I can't seem to figure out how to get access to whether an item was created or modified. The only place I see such information use is right after [a rep is written and notified](https://github.com/nanoc/nanoc/blob/2da34d8e8fc79df2c7a5fef095fbcc39a41bce0b/lib/nanoc/cli/commands/compile.rb#L69). What would be the best way to store this information so it remains in the postprocess step? Should each item handle its own state via a new attribute?

ddfreyne:
Hmm, that’s annoying.

It seems like a good idea to have `#created?` and `#modified?` on an item rep view (at least in the post-processor). I’m OK with storing that information on an item (or item rep) for now, but I imagine this might be refactored later on (and the concept of views should make this easy).

gjtorikian:
I feel like I'm pretty close to getting `#created?` and `#modified?` working. Well, I mean, these methods now *exist* and they do *something*, but something feels very dirty and wrong. I like the idea of setting the `item.rep` status after the compilation, but I'm not sure I've done it correctly. Unfortunately I'm going to have to tap out of continuing this work—I'm running up against the limitations of my knowledge of nanoc internals here. :sweat_smile:

On the bright side, it does seem like postprocessing as a whole is working.

lib/nanoc/base/views/post_compile_item_view.rb
```diff
@@ -0,0 +1,11 @@
+module Nanoc
+  class PostCompileItemView < Nanoc::ItemView
+    def created?
+      reps.select { |rep| rep.status == :created }
+    end
+
+    def updated?
+      reps.select { |rep| rep.status == :modified }
```
ddfreyne: Can you replace `:modified` with `:updated`? This way, it reflects the name of the method.

lib/nanoc/base/compilation/compiler.rb
```diff
@@ -233,8 +237,10 @@ def compile_rep(rep)
       if can_reuse_content_for_rep?(rep)
         Nanoc::Int::NotificationCenter.post(:cached_content_used, rep)
         rep.snapshot_contents = compiled_content_cache[rep]
+        rep.status = :created
```
ddfreyne: Hmm, this `:created` is incorrect. Some context:

If `#can_reuse_content_for_rep?` returns `true`, then that means that the compiled content (which was cached in the previous compilation run) can be reused without actually recompiling the item. This is used when an outdated item includes content from an item that is not outdated, and whose content can thus be reused without recompiling.

In the `ItemRepWriter`, Nanoc determines whether the resulting file exists already (`is_created`), is modified (`is_modified`), or neither. (See [these `ItemRepWriter` lines](https://github.com/nanoc/nanoc/blob/master/lib/nanoc/base/services/item_rep_writer.rb#L32-L33).)

It’d be useful to make use of the `ItemRepWriter` modified/created logic that already exists. Creating a snapshot for which a route is defined will make the item rep writer write out the compiled content. The `:last` snapshot, containing the compiled content after the entire rule is executed, is trigged by the line `executor.snapshot(rep, :last)` in the `#recalculate_content_for_rep` method below.

Perhaps the `ItemRepWriter` could update the item rep to set `item_rep.modified = is_modified`, with `is_modified` being whatever `ItemRepWriter` already calculates. (I don’t think there’s a need for a similar `created` attribute.) The advantage of this is that *only* `ItemRepWriter` determines whether content is deemed changed. In the following cases, ItemRepWriter will treat the content as not modified:

* An item that is recompiled (`#can_reuse_content_for_rep?` returns `false`), but whose content turns out to be identical after compilation

* An item whose content is reused (`#can_reuse_content_for_rep?` returns `true`)

* An item that’s not even considered for compilation (not considered outdated)

lib/nanoc/base/views/post_compile_item_view.rb
ddfreyne: See other (big) comment—these two methods might not be necessary.

lib/nanoc/base/views/post_compile_item_view.rb
```diff
@@ -0,0 +1,11 @@
+module Nanoc
+  class PostCompileItemView < Nanoc::ItemView
```
ddfreyne: :+1: for having `PostCompileItemView`. I got a little lazy and didn’t create a `PreprocessItemView` (and friends, like the collection view), but that’ll change Soon™. (See #571.)

ddfreyne:
This is looking quite good so far. Making use of the `ItemRepWriter` is the only change I’d like to see, but other than that, it’s mergeable!

gjtorikian:
Ah, thanks for [the breakdown](https://github.com/nanoc/nanoc/pull/726#discussion_r46022388). I think the naming of the class was confusing me, without doing any inspection. I believe the class is actually called when the rep is going to be written to disk. It didn't occur to me that its created/modified checks could still be useful.

I took nearly all of your suggestions. I kept the `modified` method on `PostCompileItemView`, though, since it seems like a useful helper to have.

lib/nanoc/base/entities/item_rep.rb
```diff
@@ -23,6 +23,9 @@ class ItemRep
     # @return [Enumerable<Nanoc::Int:SnapshotDef]
     attr_accessor :snapshot_defs
 
+    # @return [Boolean]
+    attr_accessor :modified
```
ddfreyne: Add `alias_method :compiled?, :compiled` for extra sweetness

gjtorikian: :+1: I like sweetness.

lib/nanoc/base/views/item_rep_view.rb
```diff
@@ -27,6 +27,11 @@ def name
       @item_rep.name
     end
 
+    # @return [Boolean]
```
ddfreyne: I like `modified` (more than `updated` which I use elsewhere) but I’m still hesitant to expose this as part of the public API just yet. I’d prefer to have it just on `PostCompileItemView` for the time being.

(I’ve been bitten by trying to expose too much before, so I’d like to avoid that!)

ddfreyne:
:+1: otherwise—looks awesome!

lib/nanoc/base/views/item_rep_view.rb
ddfreyne: In other words, could you add `# @api private` so it’s marked as private in the API docs?

gjtorikian:
Fair enough—it's your project! :fireworks:

ddfreyne:
Sweet! Will merge as soon as Travis CI gives the green light.

ddfreyne:
Thanks!

PR #730, opened 2015-11-15
Add AsciiDoc files to text_extensions

suy:


suy:
I've sent the request to master because the layout has changed, and is not a "critical" thing. Users will just add the file names to their YAML.

ddfreyne:
Awesome, thanks!

gpakosz:
Not so awesome? http://www.methods.co.nz/asciidoc/faq.html#_what_is_the_preferred_file_name_extension_for_asciidoc_files

ddfreyne:
@gpakosz Realistically, people use `.adoc` or `.asciidoc` so that text editors recognise and highlight it properly. GitHub also uses those two extensions.

ddfreyne:
FYI: http://asciidoctor.org/docs/asciidoc-recommended-practices/#document-extension

suy:
Yep, that was exactly my reasoning: On Github or Gitlab this is recognized, and the ".asc" I tend to see it as the "ascii armored GPG something".

PR #731, opened 2015-11-23
Some slight changes to .travis.yml for faster builds

gjtorikian:
* [Setting the cache to `bundler`](http://docs.travis-ci.com/user/caching/) allows the Travis container to just reuse installed gems
* [`sudo: false`](http://docs.travis-ci.com/user/workers/container-based-infrastructure/) moves the builds to the container infrastructure, which means faster builds.
* We can just limit the git clone depth to an arbitrary number (10) to make the clone a bit faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,3 +27,5 @@ cache: bundler
 
 sudo: false
 
+git:
+  depth: 10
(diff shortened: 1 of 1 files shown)


ddfreyne:
Looks good! Could you make this change to the `release-4.0.x` branch, so that the release branch also gets these improvements? I’ll be sure to merge to `master` quickly.

ddfreyne:
See #733 — I was too eager to wait :)

PR #736, opened 2015-11-27
Removed english text from link_to helper.

lvuotto:
The `link_to_unless_current` function from `link_to` helper was outputing english text. This should be avoided.


ddfreyne:
Makes sense! Could you fix the failing tests please?

rosetree:
Thanks @lvuotto!

Issue #737, opened 2015-11-27
`link_to_unless_current` attributes


lvuotto:
I think the current implementation of `link_to_unless_current` lacks the attributes flexibility of `link_to`. Giving `link_to_unless_current` the same behaviour is fairly trivial.

Has there been any discussion about this? I have no problem in submitting a pull request if the project wants to.

ddfreyne:
It makes sense for `#link_to_unless_current` to have the same behavior of `#link_to`. A PR for this would definitely be welcome!

lvuotto:
Alright, working on it. Should I take in consideration keeping the old behavoiur of adding an "active" class or should it be completely overriden by attributes hash?

ddfreyne:
Yes—retain the old behavior.

lvuotto:
And what should be the behaviour when someone set the `class` attribute? 
Append `active` always? Only if it's not present? Discard what attribute?

On 28/11/15 13:06, Denis Defreyne wrote:
> Yes—retain the old behavior.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nanoc/nanoc/issues/737#issuecomment-160314793>.
>



-- lv.

ddfreyne:
I think it makes sense to have both `class` and a new `active_class` attribute. The final HTML class would be the combination of both, if active, or just `class`.

ddfreyne:
I’ll close this issue now; see #743 for details.

PR #743, opened 2015-11-28
link_to_unless_current

lvuotto:
Fixs #737.

`attributes` hash now accepts an `:active_class` key which defaults to `'active'` and, if it links to the current page, it's merged with `attributes`' `:class` key and removed from the hash.

### DIFF
--- a/lib/nanoc/helpers/link_to.rb
+++ b/lib/nanoc/helpers/link_to.rb
@@ -88,8 +88,23 @@ module Nanoc::Helpers
       path = target.is_a?(String) ? target : target.path
 
       if @item_rep && @item_rep.path == path
+        active_class = attributes[:active_class] || 'active'
+        attributes.delete :active_class
+
+        # merge active_class with attributes' class
+        if attributes.has_key? :class then
+          attributes[:class] += " #{active_class}"
+        else
+          attributes[:class] = active_class
+        end
+
+        # Join attributes, prefixed by a space
+        attributes = attributes.reduce('') do |memo, (key, value)|
+          memo + ' ' + key.to_s + '="' + h(value) + '"'
+        end
+
         # Create message
-        "<span class=\"active\">#{text}</span>"
+        "<span#{attributes}>#{text}</span>"
       else
         link_to(text, target, attributes)
       end
--- a/test/helpers/test_link_to.rb
+++ b/test/helpers/test_link_to.rb
@@ -74,6 +74,48 @@ class Nanoc::Helpers::LinkToTest < Nanoc::TestCase
     @item = nil
   end
 
+  def test_link_to_unless_current_current_with_attributes
+    # Create item
+    @item_rep = mock
+    @item_rep.stubs(:path).returns('/foo/')
+
+    # Check
+    assert_equal(
+      '<span class="quux active" title="Foo &amp; Bar">Bar</span>',
+      link_to_unless_current('Bar', @item_rep, class: 'quux', title: 'Foo & Bar'),
+    )
+  ensure
+    @item = nil
+  end
+
+  def test_link_to_unless_current_current_with_active_class
+    # Create item
+    @item_rep = mock
+    @item_rep.stubs(:path).returns('/foo/')
+
+    # Check
+    assert_equal(
+      '<span class="foo-bar-active">Bar</span>',
+      link_to_unless_current('Bar', @item_rep, active_class: 'foo-bar-active'),
+    )
+  ensure
+    @item = nil
+  end
+
+  def test_link_to_unless_current_current_with_active_class_and_class
+    # Create item
+    @item_rep = mock
+    @item_rep.stubs(:path).returns('/foo/')
+
+    # Check
+    assert_equal(
+      '<span class="quux foo-bar-active">Bar</span>',
+      link_to_unless_current('Bar', @item_rep, class: 'quux', active_class: 'foo-bar-active'),
+    )
+  ensure
+    @item = nil
+  end
+
   def test_link_to_unless_current_not_current
     # Create item
     @item_rep = mock


ddfreyne:
The signature of the new function conflates the HTML attributes with function params which I’m not a fan of. Sadly, it seems there’s no simple way fix this, because

```ruby
def link_to_unless_current(text, target, attributes = {}, params = {})
```

will capture the attributes inside the `params` instead, and using a keyword argument e.g.

```ruby
def link_to_unless_current(text, target, attributes = {}, active_class: 'active')
```

will raise an `ArgumentError: unknown keyword` when passing in anything but `:active_class`.

A solution would be to retain the original signature:

```ruby
def link_to_unless_current(text, target, params = {})
```

and check whether `params` contains either `:active_class` or `:attributes`; if it does, fetch the attributes and active class from `params.fetch(:attributes, [])` and `params[:active_class]`, and if it does not, treat `params` as attributes.

This is getting pretty icky though, and at this point I’m wondering whether this change, given the complexity it introduces for such a tiny helper method, is worth it.

lib/nanoc/helpers/link_to.rb
```diff
@@ -88,8 +88,23 @@ def link_to_unless_current(text, target, attributes = {})
       path = target.is_a?(String) ? target : target.path
 
       if @item_rep && @item_rep.path == path
+        active_class = attributes[:active_class] || 'active'
```
ddfreyne: Use `Hash#fetch` instead, e.g. `attributes.fetch(:active_class, 'active')`

lib/nanoc/helpers/link_to.rb
```diff
@@ -88,8 +88,23 @@ def link_to_unless_current(text, target, attributes = {})
...
+        attributes.delete :active_class
```
ddfreyne: Don’t mutate `attributes`.

lib/nanoc/helpers/link_to.rb
```diff
@@ -88,8 +88,23 @@ def link_to_unless_current(text, target, attributes = {})
...
+
+        # merge active_class with attributes' class
+        if attributes.has_key? :class then
+          attributes[:class] += " #{active_class}"
+        else
+          attributes[:class] = active_class
+        end
+
+        # Join attributes, prefixed by a space
+        attributes = attributes.reduce('') do |memo, (key, value)|
```
ddfreyne: `#reduce` will be slow. Prefer appending to a string rather than creating new strings.

ddfreyne:
Some tests are failing.

gpakosz:
`link_to_unless_current` is a much used helper and based based on the reasoning we had for the `content_for` helper, I think the signature of `link_to_unless_current` should not change.

lvuotto:
For signature, what about `def link_to_unless_current(text, target, active_class: 'active', **attributes)`? It's the correct form of `def link_to_unless_current(text, target, attributes = {}, active_class: 'active')` and it will keep backwards compatibility, remove the need for `Hash#fetch`, and won't mutate `attributes`.

For the `reduce` part, I just copy-pasted the code from `link_to`. If it'll be slow, the best would be to change both functions and that escapes the scope of the issue in this PR.

For the failing tests, they're style warnings, which I'll fix and push if we come to terms about the function signature and if the patch is worthless or not.

ddfreyne:
> For signature, what about `def link_to_unless_current(text, target, active_class: 'active', **attributes)`?

This still has the issue that it conflates HTML attributes with method options. I’d rather have two different kinds of invocation: with only HTML attributes, or with method options. A combination of the two would raise an error. For example:

```ruby
# old-style invocations
link_to_unless_current('Next', next_item)
link_to_unless_current('Next', next_item, rel: 'next')

# new-style invocations
link_to_unless_current('Next', next_item)
link_to_unless_current('Next', next_item, active_class: 'current')
link_to_unless_current('Next', next_item, attributes: { rel: 'next' })
link_to_unless_current('Next', next_item, attributes: { rel: 'next' }, active_class: 'current')

# invocations that raise an error
link_to_unless_current('Next', next_item, rel: 'next', active_class: 'current')
```

Now that I think of it, it might make sense to not have `active_class` but rather `active_attributes`.

ddfreyne:
The alternative to doing all of is, is to *not* do it and follow the philosophy of only retaining the simplest possible in Nanoc.

It’s virtually impossible to provide an one-size-fits-all solution for any helper in Nanoc. Even simple ones, such as `#link_to` and `#link_to_unless_current`, already get fairly complicated due to the need for customisation.

I’d rather have a simple and not featureful implementation rather than an implementation that is complex but featureful (such as the modified `#link_to_unless_current`).

lvuotto:
I respect your opinion, I might see your point and I'll close this PR if you want to.

But before that, let me propose you a final idea: IMO the problem raises from the capability of `link_to` to generate tags with attributes, which `link_to_unless_current` lacks. What if we move this outside from `link_to` to something like `def make_tag(tag, content, **attributes)` and both `link_to` and `link_to_unless_current` just call `make_tag` with the right parameters, i.e., `link_to` will create an anchor and `link_to_unless_current` will create a span and will append "active" to `attributes`' `:class`.

ddfreyne:
@lvuotto We’d still need to have a `#link_to_unless_current` that remains backwards compatible *and* has a reasonable function signature that allows passing in existing attributes *and* an active class.

I believe it’ll lead to a significant amount of complexity regardless of the approach taken. I’m open to being challenged about this though!

(Just to be clear, I value your contributions, even if they don’t end up getting merged!)

ddfreyne:
Hi @lvuotto! Could you let me know what the status of this PR is?

lvuotto:
Hi, sorry for the delay. I haven't had time to work in this and I won't have time in the near future either. The only thing that came to my mind is a really ugly solution: setting the active class in a module variable and changing it in the preprocessor. It won't clobber html attributes with functions params and it will be backward compatible, but I think it's really, really ugly and it would do slight damage to the code readability.

Given that we couldn't agree on how to make things in way that it's both backward compatible and doesn't clobber html attributes with function params, I think we should close this PR, unless you think that the ugly solution is the way to go. If that is the case, I *think* I can do something over weekend.

ddfreyne:
I’ll close this PR now.

This is something worth revisiting when Nanoc 5 comes along (might be a while though) because that’s when breaking backwards compatibility is totally okay, and that’ll benefit the API for helpers like these.

Thanks for the time and effort you’ve put into this!

PR #744, opened 2015-11-29
Add behaviors to #content_for

### LINKED ISSUE #704: Multiple "content_for" calls should concatenate

ddfreyne:
Fixes #704.

`#content_for`, when given a block, now gains a second parameter, `:existing`, which determines what to do when existing content already exists. The three possible values are:

* `:error` (current behavior): when content already exists and is not identical, raise an error.
* `:overwrite`: overwrite the previously captured content with the newly captured content
* `:append`: append the newly captured content to the previously captured content

For example:

```rhtml
<% content_for(:stuff) do %>banana<% end %>
<% content_for(:stuff, existing: :append) do %>donkey<% end %>
```

will result in `content_for(:stuff)` containing `"bananadonkey"`

```rhtml
<% content_for(:stuff) do %>banana<% end %>
<% content_for(:stuff, existing: :error) do %>donkey<% end %>
```

will raise an exception

```rhtml
<% content_for(:stuff) do %>banana<% end %>
<% content_for(:stuff, existing: :overwrite) do %>donkey<% end %>
```

will result in `content_for(:stuff)` containing `"donkey"`

### DIFF
--- a/lib/nanoc/helpers/capturing.rb
+++ b/lib/nanoc/helpers/capturing.rb
@@ -33,17 +33,16 @@ module Nanoc::Helpers
 
       def []=(item, name, content)
         @store[item.identifier] ||= {}
-        if @store[item.identifier][name]
-          return if @store[item.identifier][name] == content
-          raise "Content_for was called twice with the same key: #{name}, this would overwrite content for the first call"
-        else
-          @store[item.identifier][name] = content
-        end
+        @store[item.identifier][name] = content
       end
 
       def [](item, name)
         @store[item.identifier] ||= {}
         @store[item.identifier][name]
+      end
+
+      def reset_for(item)
+        @store[item.identifier] = {}
       end
     end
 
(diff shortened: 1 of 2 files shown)


ddfreyne:
CC @Fjan — this  is the behavior you’re looking for.

Fjan:
Yes, thank you. I would probably still be a good idea to make :append the default at some point to make it's behaviour similar to the content_for in Rails.

ddfreyne:
That’ll have to be for Nanoc 5 (since it’s a breaking change).

Fjan:
We can be pretty confident that the current default behaviour is not used by anyone, because it would raise an exception, so I wouldn't call it a breaking change. But I'm fine with waiting until the next version.

PR #749, opened 2015-11-30
Require ItemRepCollectionView#[] param to be Symbol

### LINKED ISSUE #748: item.reps[0] returns nil rather than raising
The invocation `some_item.reps[0]` returns nil. Invoking `#[]` on `#reps` with something that is not a symbol does not make sense, and so checking whether the argument passed to `#[]` is actually a symbol would make upgrading easier.

Also see nanoc/nanoc.ws#135.

ddfreyne:
Fixes #748.


ddfreyne:
CC @theoldfellow — this will not fix the issue you’re getting, but it will raise an exception with a description that explains how to fix the issue.

ddfreyne:
Will merge this now, but a post-merge :+1: is welcome!

PR #753, opened 2015-12-03
Rule memory as path source

ddfreyne:
**Summary:** Use the rule memory to determine the (basic) path for an item rep, to prepare for adding a `#write` instruction to compilation rules.

### Current approach

The item reps get their paths (per snapshot) in the `ItemRepRouter`. The item rep router finds the appropriate routing rule, and assigns the raw path (e.g. `output/index.html`), along with the path usable for linking (e.g. `/`).

The rule memory calculator compiles the item with a “recording executor”, which does nothing but record filter/layout/… invocations.

### New approach

The rule memory calculator will determine the (basic) path for an item rep. It will do so by finding the appropriate routing rule for that item and creating an appropriate `#snapshot` rule action that includes the path of the item rep.

The item rep router is still responsible for converting this basic path (e.g. `/foo/index.html` into the path for writing to (`output/foo/index.html`) and the path for linking to (`/foo/`).

This makes it possible to add a `:path` argument to `#snapshot` in rules, so you can now write files from within the compilation rule. For example:

```ruby
compile '/*.md' do
  filter :kramdown
  snapshot :whatever, path: item.identifier.without_ext + '/index.html'
end
```

… as opposed to this:

```ruby
route '/*.md' do
  item.identifier.without_ext + '/index.html'
end

compile '/*.md' do
  filter :kramdown
end
```

Having to specify a snapshot is annoying, so there will be a convenience method `#write`:

```ruby
compile '/*.md' do
  filter :kramdown
  write item.identifier.without_ext + '/index.html'
end
```

This approach also helps pushing the rules further away from the core of Nanoc, and the rule memory might become a “processing instruction list”, which could become what Nanoc uses to determine how to process an item rep. With that in place, Nanoc could gain alternatives to the Rules file as a processing instruction source.


ddfreyne:
There are some FIXMEs and TODOs left, but I’ll tackle those in a separate PR since I believe they relate to issues that currently already exist in `master`.

PR #764, opened 2015-12-10
Fix final :pre snapshot not being generated

### LINKED ISSUE #762: :pre snapshot not generated properly on 4.1a1

ddfreyne:
The `:pre` snapshot is not generated properly (it remains “non-final”). This PR fixes that.

The `NonFinalSnapshotWithPathError` error might need to be removed. I’ll try to keep it in (it’s useful) but getting it to work properly might be hard.

Fixes #762.

### DIFF
--- a/lib/nanoc/base/entities/rule_memory.rb
+++ b/lib/nanoc/base/entities/rule_memory.rb
@@ -24,7 +24,7 @@ module Nanoc::Int
     end
 
     def add_snapshot(snapshot_name, final, path)
-      will_add_snapshot(snapshot_name)
+      will_add_snapshot(snapshot_name) if final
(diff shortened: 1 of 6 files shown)


gjtorikian:
Unfortunately, while using `compiled_content`, this still yields an error. Not sure if you wanted that fixed or not based on 

> (It might be the case that Nanoc 4.1.0a1’s behavior is in fact more correct.)

ddfreyne:
Ack—I guess that’s why it’s still in beta :).

Lesson learnt for me: when fixing a bug, try the original test case again to make sure the right bug is tackled.

I’ll reopen the original issue and look into it again.

PR #772, opened 2015-12-13
Add API for testing helpers

ddfreyne:
This is an attempt at writing an API to make testing helpers easier.

A test using this helper API would start by creating a helper context (which is used for “setting the scene” by creating items, reps, etc), and a helper (which is the reified helper module). For example:

```ruby
describe Nanoc::Helpers::Blogging do
  let(:ctx) { HelperContext.new(described_class) }
  let(:helper) { ctx.helper }
```

Next up is the subject. This calls a method on `helper`, which, as mentioned above, is an object that includes the helper module:

```ruby
  describe '#articles' do
    subject { helper.articles }
```

To set up the context, we use methods like `#create_item` on `ctx` (the helper context). We have to use `#let!` rather than `#let` (notice the bang) to prevent them from being lazily evaluated. In this example, one item is created:

```ruby
    let!(:item_a) do
      ctx.create_item('blah', { kind: 'item' }, '/0/')
    end
```

Lastly, there’s the actual test, which calls the helper as you would normally:

```ruby
    it 'returns the two articles' do
      expect(subject.map(&:identifier)).to match_array(['/1/', '/2/'])
    end
```

### DIFF
--- a/spec/nanoc/helpers/blogging_spec.rb
+++ b/spec/nanoc/helpers/blogging_spec.rb
@@ -1,136 +1,50 @@
 describe Nanoc::Helpers::Blogging do
-  let(:mod) do
-    Class.new(Nanoc::Int::Context) do
-      include Nanoc::Helpers::Blogging
-    end.new(assigns)
-  end
-
-  let(:assigns) do
-    { items: [] }
-  end
-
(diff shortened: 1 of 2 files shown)


ddfreyne:
I’ll merge this now, even though the API is still considered experimental/private. Future work includes migrating all existing tests to specs using this new API.

PR #779, opened 2015-12-15
Do not prune other snapshots

### LINKED ISSUE #776: Files generated by snapshots are pruned away
In Nanoc 4.1.0rc2, a compilation rule with a `#snapshot` call along with a routing rule matching that snapshot, will cause the item rep to be written, but pruned afterwards.

ddfreyne:
Fixes #776.


ddfreyne:
The pruner isn’t terribly cleanly written and I find that just the new regression test is okay for now.

PR #780, opened 2015-12-17
Extract match opts out as a constant

gjtorikian:
While mucking about I noticed this could probably be made a class-level constant rather than recomputing on each call.


ddfreyne:
Neat!

Mucking about is the way most of the Nanoc code got written. It’s pretty effective!

PR #789, opened 2015-12-31
Avoid double calculation of checksum

RubenVerborgh:
The new checksum of objects is currently calculated twice:

1. for `checksums_available?`
2. for `checksums_identical?`

Since checksum calculation can be costly, this commit memoizes its calculation.

This speeds up the compilation of my site with 20% in case of no changes.


ddfreyne:
Thanks!

PR #791, opened 2016-01-01
Updated copyright to 2016

aqnouch:


ddfreyne:
Thanks! I’ve taken care of this in c0e16c03d8250641ee9521088507e906eb45395b, which is on the latest release branch rather than `master`, where I’ve also updated the copyright year in the executable itself.

PR #792, opened 2016-01-01
Add new helper specs

ddfreyne:
This replaces some of the old tests with new, cleaner specs.

Hopefully, once everything is converted into nice specs, the helper context can become part of the public API. That’s not the goal of this PR, however.


ddfreyne:
This does not replace all helper tests, but the existing ones are too high-level or complicated to rewrite in a nice way. Might do this later.

Issue #790, opened 2015-12-31
Speed up checksum calculation for items


RubenVerborgh:
I noticed that roughly 80% of my site's compilation time is due to checksum calculation. While #789 eliminates some of that time, it still weighs on the speed of nanoc.

I think that (some) items can be smarter about their checksum calculation. For instance, an item from the filesystem could generate its checksum based on the SHA1 hash of the file contents, rather than having to parse the entire file and calculating a SHA1 hash of all fields, as is done now. Other data sources could calculate checksums even faster. For instance, the [BibTeX datasource](https://github.com/RubenVerborgh/nanoc-bibtex) could generate one checksum per .bib file, instead of having to generate a checksum for each of the individual entries in that file.

I would like to implement this idea and submit a pull request, but I'd appreciate some guidance on how to implement this. In particular, I'm wondering about the following:

- Currently, `Nanoc::Int::Checksummer` is responsible for its checksumming all other classes. However, this makes that component rather omniscient, in contrast with the principles of object-oriented design. Wouldn't it be good that each entity is responsible for checksumming itself?
  - If so, this would mean we need to extend basic classes like `String` with a checksum method. This might or might not be desired.
  - An alternative is to implement `Checksummer` such that it first checks whether the object supports a `calc_checksum` method and if so, calls it.
- I see two ways in which a datasource could work:
  - Either it would generate a field such as `:checksum`, which the `checksum` method then uses if it exists.
  - Otherwise, it would directly implement the `calc_checksum` method on the item it returns, either by monkey patching or subclassing. However, I wonder to what extent subclassing is desired, given that `Item` has moved to the Nanoc internals.
    - On the theme of subclassing and speed: having subclasses of `Item` would also allow us to have dedicated methods to access properties, which would allow lazy loading. Alternatively, the `new_item` method could accept a block as parameter, which is only executed when the properties are really needed.
- What does a `calc_checksum` look like? Does it take a digest as argument, which it then updates? Or does it take no arguments?

ddfreyne:
> Other data sources could calculate checksums even faster.

That’s an interesting idea! The advantage of the current approach to generating checksums is that no-op changes (e.g. formatting changes) won’t cause a change in the checksum, but I guess that’s really just an edge case.

> Currently, `Nanoc::Int::Checksummer` is responsible for its checksumming all other classes. However, this makes that component rather omniscient, in contrast with the principles of object-oriented design. Wouldn't it be good that each entity is responsible for checksumming itself?

[...]

PR #793, opened 2016-01-01
Add checksum parameter to Document

RubenVerborgh:
As suggested in #790, this pull request allows a `Document` to specify a pre-calculated checksum. This gives data sources the ability to determine a checksum for the items and layouts they create.
This pull request:

- adds a `checksum` parameter to the initializer of `Document` (as well as `new_item` and `new_layout`)
- modifies `Checksummer` to use the `checksum` on `Item` and `Layout` if present
- modifies the filesystem datasource to set the size/modification time as checksum

This realizes a major compilation speed gain for my website (from 11s to 3s if nothing is changed), and paves the way for lazy loading of items, which could bring down compilation time even further.

### DIFF
--- a/lib/nanoc/data_sources/filesystem.rb
+++ b/lib/nanoc/data_sources/filesystem.rb
@@ -61,7 +61,8 @@ module Nanoc::DataSources
           elsif is_binary && klass == Nanoc::Int::Layout
             raise "The layout file '#{content_filename}' is a binary file, but layouts can only be textual"
           else
-            meta, content_or_filename = parse(content_filename, meta_filename, kind)
+            meta, content_or_filename, meta_raw = parse(content_filename, meta_filename, kind)
+            checksum_data = "content=#{content_or_filename},meta=#{meta_raw}"
           end
 
           # Get attributes
@@ -105,7 +106,7 @@ module Nanoc::DataSources
             end
 
           # Create object
-          res << klass.new(content, attributes, identifier)
+          res << klass.new(content, attributes, identifier, checksum_data: checksum_data)
         end
       end
 
@@ -207,8 +208,8 @@ module Nanoc::DataSources
     end
 
     # Parses the file named `filename` and returns an array with its first
-    # element a hash with the file's metadata, and with its second element the
-    # file content itself.
+    # element a hash with the file's metadata, its second element the
+    # file content itself, and its third element the metadata content.
     def parse(content_filename, meta_filename, _kind)
       # Read content and metadata from separate files
       if meta_filename
@@ -249,7 +250,7 @@ module Nanoc::DataSources
       content = pieces[4]
 
       # Done
-      [meta, content]
+      [meta, content, meta_raw]
     end
 
     class InvalidMetadataError < Nanoc::Error
--- a/lib/nanoc/base/checksummer.rb
+++ b/lib/nanoc/base/checksummer.rb
@@ -91,14 +91,18 @@ module Nanoc::Int
         when Nanoc::Int::BinaryContent
           update(Pathname.new(obj.filename), digest)
         when Nanoc::Int::Item, Nanoc::Int::Layout
-          digest.update('content=')
-          update(obj.content, digest)
+          if obj.checksum_data
+            digest.update('checksum_data=' + obj.checksum_data)
+          else
+            digest.update('content=')
+            update(obj.content, digest)
 
-          digest.update(',attributes=')
-          update(obj.attributes, digest, visited + [obj])
+            digest.update(',attributes=')
+            update(obj.attributes, digest, visited + [obj])
 
-          digest.update(',identifier=')
-          update(obj.identifier, digest)
+            digest.update(',identifier=')
+            update(obj.identifier, digest)
+          end
         when Nanoc::ItemWithRepsView, Nanoc::ItemWithoutRepsView, Nanoc::LayoutView, Nanoc::ConfigView, Nanoc::IdentifiableCollectionView
           update(obj.unwrap, digest)
         else
--- a/lib/nanoc/base/entities/document.rb
+++ b/lib/nanoc/base/entities/document.rb
@@ -11,15 +11,21 @@ module Nanoc
       # @return [Nanoc::Identifier]
       attr_accessor :identifier
 
+      # @return [String, nil]
+      attr_accessor :checksum_data
+
       # @param [String, Nanoc::Int::Content] content
       #
       # @param [Hash] attributes
       #
       # @param [String, Nanoc::Identifier] identifier
-      def initialize(content, attributes, identifier)
+      #
+      # @param [String, nil] checksum_data Used to determine whether the document has changed
+      def initialize(content, attributes, identifier, checksum_data: nil)
         @content = Nanoc::Int::Content.create(content)
         @attributes = attributes.__nanoc_symbolize_keys_recursively
         @identifier = Nanoc::Identifier.from(identifier)
+        @checksum_data = checksum_data
       end
 
       # @return [void]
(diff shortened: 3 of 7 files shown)


RubenVerborgh:
Note that abebd781db309d0c26c3517af50a1de738ac63bd is redundant after 871cd9a6e25d2241345eea956940f24aa7ec9288; however, I left it in for comparison (especially with regard to the tests).

RubenVerborgh:
Flagged for style issues. Let me fix that :smile:

RubenVerborgh:
Also, the [code to generate a checksum from a path](https://github.com/nanoc/nanoc/pull/793/files#diff-ae21b65adeb38854fa5b4568014c8f0dR68) is kind of a duplicate of [the code of the checksummer for pathnames](https://github.com/nanoc/nanoc/blob/6404b23ecfc26e8a42ac9d46b4b25e3a3e351c84/lib/nanoc/base/checksummer.rb#L74). I did not want to create a dependency between a datasource and the checksummer; while this might be okay for the default _filesystem_ datasource, external datasources would not want to depend on the internal checksummer. (Would also be an argument to have pathnames calculate their own checksum, but we've discussed that already :wink:)

ddfreyne:
This looks good! Some remarks:

* I’m not in favour of using the file size and mtime to detect modifications. It’s not accurate enough—it’s trivial to generate two different files with the same checksum. The content and attributes (as strings) are already loaded, and generating a checksum from a string is quite fast (SHA-1 is intended to be fast). For binary items, this isn’t as straightforward, and file size + mtime will have to do.

  I’m not saying no to using mtime at all, but for a checksum, I don’t think it suffices. (Idea: only calculate a full checksum when the mtime appears to indicate that the item is outdated—but I believe such a change is out of scope for this PR and would require quite a bit more (architectural) rework.)

* The checksummer is part of a private API, but I have no issues with the filesystem data source using that API. It is part of the private API because I don’t want to expose it outside of Nanoc (and have to maintain it as part of the public API).

RubenVerborgh:
- I wasn't too sure either, hence I left 289f6f38c5bac50e66abdcd42710bfa657c35b72 in. I'll adjust the range of this pull request to not include 8c72d5159151a8f980d7ddd103fa5e5277fe397a anymore.

- Alright, but it's not necessary anymore then if the mtime is not used.

ddfreyne:
> Content only needs to be loaded when the checksum is different.

Not entirely correct: the checksum being different is a necessary but not sufficient condition for content to be loaded. For instance, an item (whose checksum is identical) that includes content from another item (whose checksum is different) will need to be recompiled even though its checksum hasn’t changed.

The idea of only loading data when necessary is nonetheless quite interesting and is something I plan to tackle in Nanoc at some point in the future.

lib/nanoc/data_sources/filesystem.rb
```diff
@@ -249,7 +251,7 @@ def parse(content_filename, meta_filename, _kind)
       content = pieces[4]
 
       # Done
-      [meta, content]
+      [meta, content, meta_raw]
```
ddfreyne: I don’t think `meta_raw` is used.

lib/nanoc/data_sources/filesystem.rb
```diff
@@ -58,10 +58,12 @@ def load_objects(dir_name, kind, klass)
           if is_binary && klass == Nanoc::Int::Item
             meta                = (meta_filename && YAML.load_file(meta_filename)) || {}
             content_or_filename = content_filename
+            checksum = nil
           elsif is_binary && klass == Nanoc::Int::Layout
             raise "The layout file '#{content_filename}' is a binary file, but layouts can only be textual"
           else
-            meta, content_or_filename = parse(content_filename, meta_filename, kind)
+            meta, content_or_filename, meta_raw = parse(content_filename, meta_filename, kind)
+            checksum = "meta=#{meta_raw},content=#{content_or_filename}"
```
ddfreyne: Pass this string to `Nanoc::Int::Checksummer.calc` rather than using it as-is (otherwise the checksum will be enormous).

lib/nanoc/data_sources/filesystem.rb
ddfreyne: nm—it is!

ddfreyne: Can you update the `#parse` documentation to reflect the changed return value?

ddfreyne:
:+1: apart from comments

lib/nanoc/data_sources/filesystem.rb
RubenVerborgh: The only consumer of the checksum is `Nanoc::Int::Checksummer.calc`. Does it really need to be sent through the checksummer a second time?

RubenVerborgh:
> > Content only needs to be loaded when the checksum is different.
> Not entirely correct: the checksum being different is a necessary but not sufficient condition for content to be loaded.

Yeah, I cut a corner in writing that down. What I meant is: content only needs to be loaded when a reason is found to recompile, since determining that reason no longer requires the item contents.

> The idea of only loading data when necessary is nonetheless quite interesting and is something I plan to tackle in Nanoc at some point in the future.

I think it could be as simple as allowing `new_item` to accept a block and a checksum, but I might be oversimplifying. Can I make an issue to discuss that?

lib/nanoc/data_sources/filesystem.rb
ddfreyne: Hmm, good point. Perhaps renaming the parameter to `checksum_data`, making it clear that it is not the checksum itself, is enough to clear up the confusion.

ddfreyne:
> Yeah, I cut a corner in writing that down. What I meant is: content only needs to be loaded when a reason is found to recompile, since determining that reason no longer requires the item contents.

Yup.

> I think it could be as simple as allowing `new_item` to accept a block and a checksum, but I might be oversimplifying. Can I make an issue to discuss that?

I have a few ideas regarding optimisations like these, and I’d rather write down these thoughts in a proper RFC first. (Can’t seem to find the time, though.)

lib/nanoc/data_sources/filesystem.rb
RubenVerborgh: Will do that!

RubenVerborgh:
Okay, keep me updated when that RFC goes out :smile:

Finishing this PR now.

RubenVerborgh:
All done.

ddfreyne:
Looks good, thanks!

ddfreyne:
Some follow-up discussion:

* Using mtime+file size as an initial checksum, and only calculating the full checksum if necessary (and only then loading content) is pointless, because when the mtime/file size checksum is different, the item will need to be recompiled and that means loading the content anyway. Additionally, the full checksum will need to be calculated for this item anyway, because it needs to be stored so that it can be used in future compilations. (The previous checksum could be reused rather than recalculated, but I’m unsure whether the speedup would be large enough to make it worth optimise this.)

* Not loading content, or lazy-loading content, made much more sense in previous versions of Ruby, but since Ruby as of recently has a generational GC, the performance gains of keeping memory usage low are not nearly as big anymore.

RubenVerborgh:
Lazy loading still makes sense, not for memory reasons, but to avoid the overhead of parsing the file and its metadata. The checksum can be made without parsing – only when the checksum is different, the file actually needs to be parsed.

ddfreyne:
I’d argue that in most cases, parsing the metadata doesn’t impose enough of an overhead to try to avoid doing it—but I haven’t profiled this extensively to be certain of that.

RubenVerborgh:
I can give you a quick number: a site with a BibTeX datasource (± 150 items in my use case), compilation  time was 12s before #790 and #793. Now it is down to 3s. Skipping field loading for the 150 items brings compilation time further down to 2s.

ddfreyne:
This is the “nothing changed” case, correct?

RubenVerborgh:
Yes.

PR #794, opened 2016-01-03
Lazy-load items

RubenVerborgh:
This pull request allows documents to lazily load attributes and content, as discussed in nanoc/rfcs#3.

### To do
- [x] update `new_item` documentation
- [ ] implement support in filesystem datasource


RubenVerborgh:
Tested this with my BibTeX datasource, and I'm seeing a compilation speed gain already.

Two questions for @ddfreyne:

- Am I correct to implement the laziness for content in `Content`, or should this be in `Document`? My reasoning for changing `Content`: `new_item` directly uses `Content.create`.
- I can't wrap my head around why I [can modify `@attributes` on a frozen document](https://github.com/nanoc/nanoc/commit/5ef8dcb2d1c161345f392ccd44d16c25f8d63eac#diff-ac36bae0fe8ca23e003b3cd2345ced8eR11), but [need this trickery](https://github.com/nanoc/nanoc/commit/cc7a68adcd11787f4752c97bd66e7198939c04e4#diff-bb929efbc4546edde5dac9db28e3a565R62) to modify `string` on frozen content. Any clue?

ddfreyne:
The frozenness and symbolisation handling feels a little awkward, and I think it could do with some more abstraction. Here’s an idea:

```ruby
class LazyValue
  def initialize(thing)
    @thing = thing
  end

  def value
    @value ||= transform(raw_value)
  end

  def raw_value
    @raw_value ||=
      if @thing.respond_to?(:call)
        @thing.call
      else
        @thing
      end
  end

  def transform(x)
    x
  end

  def freeze
    value.freeze
  end
end

class LazyAttributesValue < LazyValue
  def transform(x)
    x.__nanoc_symbolize_keys_recursively
  end

  def freeze
    value.__nanoc_freeze_recursively
  end
end

class LazyContentValue < LazyValue
end
```

That code could be used like this (pseudocode):

```ruby
class TextualContent < Content
  def string
    @string_holder.value
  end
end

class Document
  def attributes
    @attributes_holder.value
  end
end
```

Untested, but hopefully this gets the idea across! This approach is purely functional and easier to use/understand.

Not sure about the “holder” name though.

ddfreyne:
> Am I correct to implement the laziness for content in `Content`, or should this be in `Document`?

Seems correct!

> Does the laziness handling with the holder seem alright? This was needed for freezing.

It’s confusing, and it’s not DRY, so it’s a little tricky to tell. I think abstracting the laziness handling away (e.g. using the suggestion above) could make this easier.

RubenVerborgh:
Your suggestion of `LazyValue` also crossed my mind, especially if this becomes a common thing (I was still missing the `transform` idea though). Where should I put this auxiliary class?

ddfreyne:
`lib/nanoc/base/entities/lazy_value.rb` would work! Put it inside `Nanoc::Int` though, to make it clear this is for internal use only.

RubenVerborgh:
Alright, will do tomorrow. BTW, any reason to prefer `respond_to?(:call)` over `is_a?(Proc)`?

ddfreyne:
I prefer  prefer `respond_to?(:call)` over `is_a?(Proc)` because it makes duck typing easier—not just procs can yield values. This is also in line with how Rack apps are built (the contract there is to respond to `#call`.)

RubenVerborgh:
Implemented `Nanoc::Int::LazyValue`. I did not entirely follow your suggestion, as your `freeze` method requires the value to be computed, which kind of voids the purpose of being lazy. I implemented it with a hash that only becomes frozen when the value is actually requested.

RubenVerborgh:
Don't merge this yet; even though the tests pass, there is a very strange bug which occurs when compiling a site. At a certain point in time, I end up with a `TextualContent` whose `@string` property, [explicitly initialized to be a `LazyValue`](https://github.com/nanoc/nanoc/commit/8737d78fa91722ed082c2ffea10a5fd6edb1d19f#diff-bb929efbc4546edde5dac9db28e3a565R67) has type `String`. This is not possible if the object is constructed normally. I verified that other constructor steps are skipped also.

So, are there any places where `TextualContent` is not constructed in a regular way?

RubenVerborgh:
…and now the bug has disappeared magically. I wonder, could this happen when content is somehow resurrected from cache?

RubenVerborgh:
Careful: old temp files seem to be affected by this change. PStore tries resurrecting the item with `@string` as a string value, which is not the case anymore. Simply deleting the `tmp` folder works. In general, however, it seems that nanoc should be careful reading temp files.

ddfreyne:
Hmm, loading cached data might be a problem. I’ll look into whether or not this has a significant impact.

The tmp store has a version number; bumping it would be good (it’ll invalidate previous data).

RubenVerborgh:
Alright, definitely needs to be bumped.

ddfreyne:
Bump the version of the compiled content cache by changing the number argument in the initializer. In this example, change it from 1 to 2:

```ruby
  class CompiledContentCache < ::Nanoc::Int::Store
    def initialize
      super('tmp/compiled_content', 1)
```

It’s a good idea to not only bump the version, but also implement `#marshal_dump` and `#marshal_load` on `TextualContent` (and perhaps `BinaryContent`, for consistency) so that it stores/loads the string value:

```ruby
      def marshal_dump
        [@filename, string]
      end

      def marshal_load(array)
        @filename = array[0]
        @string = Nanoc::Int::LazyValue.new(array[1])
      end
```

RubenVerborgh:
I implemented `marshal_dump` and `marshal_load` for `TextualContent`. Did not do this for `BinaryContent`, as it complexifies without actually changing behavior (but I can still add this is you like).

Support for lazy loading is now complete.

Shall I also implement this for the `Filesystem` source? In this PR or another? And do you have any advice/preference for tackling this, given that the `load_objects` code is already quite complex?

ddfreyne:
I believe updating the Filesystem data source is worth it, although I’d do it in a separate PR. (The code for the filesystem data source is quite complex, and perhaps it warrants a refactoring first. I think it’s easier if I do that, but you’re welcome to give it a shot too.)

ddfreyne:
Looks good! Thanks :)

RubenVerborgh:
Thanks—agree on the refactoring, it seems best I leave this up to you.

PR #799, opened 2016-01-13
Remove env from crash.log

ddfreyne:
CC @gjtorikian


ddfreyne:
Wrong base branch; will redo.

Issue #798, opened 2016-01-12
Remove environment variables from crash.log?


gjtorikian:
Howdy! I just noticed that the crash.log dump file spits out a user's entire environment. I'm concerned that, in the event that a user has sensitive information (GitHub tokens, AWS tokens, etc), these variables will be leaked. 

Poking around at random issues in this repo, I see a few places where people just straight up paste the log: https://github.com/nanoc/nanoc/issues/199, https://github.com/nanoc/nanoc/issues/405.

Would it be possible to just omit this information?

/cc @mastahyeti

ddfreyne:
Yup, totally makes sense. Having the environment available in `crash.log` wasn’t as useful as I imagined in the first place, and when it’s necessary, we can still get hold of it by other means.

PR: #800

ddfreyne:
Fixed in #800. This’ll be part of the 4.1.2 release.

PR #800, opened 2016-01-13
Remove env from crash.log

ddfreyne:
The environment can contain sensitive information, and so it shouldn’t be included in `crash.log` by default.

CC @gjtorikian 

See #798.


ddfreyne:
The error handler is lacking in detailed tests, hence no change in specs/tests.

gjtorikian:
:heart:

PR #802, opened 2016-01-16
Refactor filesystem data source

ddfreyne:
This is in preparation of returning content/attributes wrapped in a `LazyValue`. See #794.


ddfreyne:
Probably still a work in progress, but has made enough progress to be merged.

PR #818, opened 2016-02-05
Check type in #link_to argument

### LINKED ISSUE #816: Let #link_to fail when passed an identifier
The following should fail explicitly because of an unexpected type being passed to `#link_to`:

```erb
<%= link_to i[:title], i.identifier %>
```

ddfreyne:
Fixes #816.


ddfreyne:
Wrong branch.

PR #835, opened 2016-03-05
Fix show-data crash

### LINKED ISSUE #833: nanoc show-data crashes
Steps to reproduce:

```
nanoc create-site FooBar
cd FooBar
nanoc show-data
```

crashes with the following error :

```
NoMethodError: undefined method `reps' for <Nanoc::Int::Item identifier="/index.html">:Nanoc::Int::Item
```

ddfreyne:
Fixes #833.

This PR has a fix and a regression test, but I’d like to take the time to write a proper spec for the `show-data` command. That command has been around for a while and has always “just worked”—until now, apparently, and I’d like to ensure that it keeps on working.

### DIFF
--- a/lib/nanoc/cli/commands/show-data.rb
+++ b/lib/nanoc/cli/commands/show-data.rb
@@ -40,7 +40,7 @@ module Nanoc::CLI::Commands
     def sorted_reps_with_prev(items)
       prev = nil
       items.sort_by(&:identifier).each do |item|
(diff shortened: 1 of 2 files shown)


ddfreyne:
I’m OK with only having a regression test for now.

The CLI commands have generally rather bad code coverage (and the way they’re loaded makes it impossible for the code coverage tool to detect them), so I’d like to tackle that in a separate PR.

PR #843, opened 2016-03-26
Replace LazyValue#transform with #map

ddfreyne:
This removes `#transform` and provides a more composable `#map`:

```ruby
lv = Nanoc::Int::LazyValue.new(-> { sleep 1; 'Hello' })
lv_uppercase = lv.map(&:upcase)
p lv_uppercase.value
p lv_uppercase.value
```

This eliminates the need for subclassing `LazyValue` to specify custom transformations, and it allows chaining transformations.

Next up: updating the filesystem data source to make use of these new lazy values.

CC @RubenVerborgh — this will likely break your customisations since I believe they depend on `#transform`

### DIFF
--- a/lib/nanoc/base/entities/document.rb
+++ b/lib/nanoc/base/entities/document.rb
@@ -25,7 +25,7 @@ module Nanoc
       # @param [String, nil] checksum_data Used to determine whether the document has changed
       def initialize(content, attributes, identifier, checksum_data: nil)
         @content = Nanoc::Int::Content.create(content)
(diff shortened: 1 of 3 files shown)


RubenVerborgh:
@ddfreyne Thanks for notifying me, but `transform` was only used in  `LazyAttributesValue` of `Document` within Nanoc itself, not in my code.

ddfreyne:
Good—that’s part of a private API anyway ;)

PR #853, opened 2016-05-04
add missing contributor

ccmtaylor:


gpakosz:
Denis is the author, and by definition must not appear in the contributors list.

ddfreyne:
Arguably, I’ve contributed the most to Nanoc, so it makes sense for me to be in there. (I guess it would make sense to change the copyright line to “Nanoc © 2007–2016 Denis Defreyne and contributors” though.)

Build broke on old Ruby versions because of a dependency. Will fix elsewhere.

Issue #676, opened 2015-07-11
--config option on commandline


ddfreyne:
Having a `--config` option on the command line will make it possible to have different environments.

ddfreyne:
It will also be necessary to specify the location of the `tmp` directory—something which I assumed was already implemented, but I was mistaken.

Fjan:
IMHO a minimal implementation of this comprises these things:

1. A global variable that indicates the environment. 
2. A way to switch the output directory for each environment. 
3. A way to switch the tmp directory for each environment

It's actually already possible to do all three with the current nanoc (we use it successfully):

1. We simply use something like ENV['NANOC_ENVIRONMENT'], and change the rules in the Rules files with that, but it's probably nicer if something like  ::Nanoc.env existed
2. The Rules files accepts relative paths, so it's possible to write an output file to 'output/../output_production/...' based on the environment variable. A --config option to use a different yaml file would be nice, but alternatively, we could do it like Rails and run nanoc.yaml file through Erb, so you can change the directories based on the environment variable
3. It's a one line monkey patch to the Nanoc::Store class to change the name of tmp dir depending on the environment variable, but it would of course be nicer to have a config option for it too

We wrap the relevant nanoc compile command in custom nanoc commands and rake tasks to set the right ENV variable beforehand. It works well actually, but it would be great if we can do it without monkey patching.

Fjan:
So a --config option is one way to do it, but an --env option that gets picked up in the nanoc.yaml like rails does it may be even nicer, because there are a bunch of other things you may want to change based on the environment than just the directories. A few examples from our use case:
- The production version uses the CDN for images, the locally compiled output uses local images
- The production version has analytics snippets, the local version does not
- The local version has  output from the spell check filter in it, the production version does not
- Pages that are still being worked on are hidden in the production environment

bburton:
Agree with @Fjan on --env/-e option for specifying an environment name. Those options seem to be commonly used in many Ruby apps including compass which is used with Nanoc by many people. But then there needs to be a way to load the environment specific configuration. It would probably be best to always load `nanoc.yaml` and if -e/--env is specified or the `NANOC_ENV` environment variable is defined, then load the specified environment specific name from a file such as `nanoc-production.yaml` if "production" was specified either way.

[...]

PR #859, opened 2016-05-31
Support nanoc environments

barraq:
# Description

Introduce support for nanoc environment and addresses https://github.com/nanoc/nanoc/issues/676.

- nanoc command is updated to include `--env` argument
- site config is overriden using the active environment if any and fallback to 'default'
- if environments is provided, `tmp` directory is updated to `tmp/{{env_name}}`

Setting environments is done in nanoc.yaml using the `environments` property.
Example usage is:

```
output_dir: output

environments:
  default: &default
    base_url: ...
    ...
  development:
    <<: *default
    base_url: ...
  production:
    <<: *default
    base_url: ...
  yet_another_env:
    <<: *default
    base_url: ...
    output_dir: build
```

Selecting working environment can be done:
- using environment variable `NANOC_ENVIRONMENT`
- using `nanoc --env=[<value>]`

# Progress

- [x] Main Nanoc command (cli/commands/nanoc.rb) is updated to include `--env` option
- [x] ```Nanoc::Int::Configuration``` is updated to provide `with_environment(env = nil)`
- [x] unless define in environment, output_dir is set as `{{output_dir}}/{{env_name}}`, if `environments` is not defined then `output_dir` remains unchanged
- [x] if environments is provided, `tmp` directory is updated to `tmp/{{env_name}}`
- [x] New tests are added

### DIFF
--- a/lib/nanoc/base/entities/configuration.rb
+++ b/lib/nanoc/base/entities/configuration.rb
@@ -68,9 +68,6 @@ module Nanoc::Int
       # Load given environment configuration
       env_config = @wrapped[ENVIRONMENTS].fetch(env.to_sym, {})
 
-      # Handle specific properties
-      env_config[:output_dir] ||= File.join(@wrapped[:output_dir].to_s, env.to_s)
-
       self.class.new(@wrapped, env).merge(env_config)
     end
 
--- a/lib/nanoc/base/entities/configuration.rb
+++ b/lib/nanoc/base/entities/configuration.rb
@@ -33,11 +33,15 @@ module Nanoc::Int
       string_pattern_type: 'glob',
     }.freeze
 
+    attr_reader :env
+
     contract Hash => C::Any
     # Creates a new configuration with the given hash.
     #
     # @param [Hash] hash The actual configuration hash
-    def initialize(hash = {})
+    # @param [Symbol, String] env The active environment for this configuration
+    def initialize(hash = {}, env = nil)
+      @env = env
       @wrapped = hash.__nanoc_symbolize_keys_recursively
     end
 
@@ -49,6 +53,21 @@ module Nanoc::Int
       end
 
       self.class.new(new_wrapped)
+    end
+
+    def with_environment(env = nil)
+      return self unless @wrapped.key? :environments
+
+      # Set active environment
+      env ||= ENV.fetch 'NANOC_ENVIRONMENT', :default
+
+      # Load given environment configuration
+      env_config = @wrapped[:environments][env.to_sym] || {}
+
+      # Handle specific properties
+      env_config[:output_dir] ||= File.join(@wrapped[:output_dir].to_s, env.to_s)
+
+      self.class.new(@wrapped, env).merge(env_config)
     end
 
     contract C::None => Hash
@@ -86,12 +105,12 @@ module Nanoc::Int
 
     contract C::Or[Hash, self] => self
     def merge(hash)
-      self.class.new(@wrapped.merge(hash.to_h))
+      self.class.new(@wrapped.merge(hash.to_h), @env)
     end
 
     contract C::Any => self
     def without(key)
-      self.class.new(@wrapped.reject { |k, _v| k == key })
+      self.class.new(@wrapped.reject { |k, _v| k == key }, @env)
     end
 
     contract C::Any => self
--- a/lib/nanoc/base/entities/configuration.rb
+++ b/lib/nanoc/base/entities/configuration.rb
@@ -33,13 +33,17 @@ module Nanoc::Int
       string_pattern_type: 'glob',
     }.freeze
 
+    # @return [String, nil] The active environment for the configuration
     attr_reader :env
+
+    # Configuration environments property key
+    ENVIRONMENTS = :environments
 
     contract Hash => C::Any
     # Creates a new configuration with the given hash.
     #
     # @param [Hash] hash The actual configuration hash
-    # @param [Symbol, String] env The active environment for this configuration
+    # @param [String, nil] env The active environment for this configuration
     def initialize(hash = {}, env = nil)
       @env = env
       @wrapped = hash.__nanoc_symbolize_keys_recursively
@@ -55,14 +59,14 @@ module Nanoc::Int
       self.class.new(new_wrapped)
     end
 
-    def with_environment(env = nil)
-      return self unless @wrapped.key? :environments
+    def with_environment
+      return self unless @wrapped.key?(ENVIRONMENTS)
 
       # Set active environment
-      env ||= ENV.fetch 'NANOC_ENVIRONMENT', :default
+      env = @env || ENV.fetch('NANOC_ENV', 'default')
 
       # Load given environment configuration
-      env_config = @wrapped[:environments][env.to_sym] || {}
+      env_config = @wrapped[ENVIRONMENTS].fetch(env.to_sym, {})
 
       # Handle specific properties
       env_config[:output_dir] ||= File.join(@wrapped[:output_dir].to_s, env.to_s)
--- a/lib/nanoc/base/entities/configuration.rb
+++ b/lib/nanoc/base/entities/configuration.rb
@@ -34,7 +34,7 @@ module Nanoc::Int
     }.freeze
 
     # @return [String, nil] The active environment for the configuration
-    attr_reader :env
+    attr_reader :env_name
 
     # Configuration environments property key
     ENVIRONMENTS = :environments
@@ -43,9 +43,9 @@ module Nanoc::Int
     # Creates a new configuration with the given hash.
     #
     # @param [Hash] hash The actual configuration hash
-    # @param [String, nil] env The active environment for this configuration
-    def initialize(hash = {}, env = nil)
-      @env = env
+    # @param [String, nil] env_name The active environment for this configuration
+    def initialize(hash = {}, env_name = nil)
+      @env_name = env_name
       @wrapped = hash.__nanoc_symbolize_keys_recursively
     end
 
@@ -63,12 +63,12 @@ module Nanoc::Int
       return self unless @wrapped.key?(ENVIRONMENTS)
 
       # Set active environment
-      env = @env || ENV.fetch('NANOC_ENV', 'default')
+      env_name = @env_name || ENV.fetch('NANOC_ENV', 'default')
 
       # Load given environment configuration
-      env_config = @wrapped[ENVIRONMENTS].fetch(env.to_sym, {})
+      env_config = @wrapped[ENVIRONMENTS].fetch(env_name.to_sym, {})
 
-      self.class.new(@wrapped, env).merge(env_config)
+      self.class.new(@wrapped, env_name).merge(env_config)
     end
 
     contract C::None => Hash
@@ -106,12 +106,12 @@ module Nanoc::Int
 
     contract C::Or[Hash, self] => self
     def merge(hash)
-      self.class.new(@wrapped.merge(hash.to_h), @env)
+      self.class.new(@wrapped.merge(hash.to_h), @env_name)
     end
 
     contract C::Any => self
     def without(key)
-      self.class.new(@wrapped.reject { |k, _v| k == key }, @env)
+      self.class.new(@wrapped.reject { |k, _v| k == key }, @env_name)
     end
 
     contract C::Any => self
--- a/lib/nanoc/base/entities/configuration.rb
+++ b/lib/nanoc/base/entities/configuration.rb
@@ -41,12 +41,12 @@ module Nanoc::Int
     NANOC_ENV = 'NANOC_ENV'.freeze
     NANOC_ENV_DEFAULT = 'default'.freeze
 
-    contract Hash, C::Maybe[String] => C::Any
+    contract C::KeywordArgs[hash: C::Optional[Hash], env_name: C::Maybe[String]] => C::Any
     # Creates a new configuration with the given hash.
     #
     # @param [Hash] hash The actual configuration hash
     # @param [String, nil] env_name The active environment for this configuration
-    def initialize(hash = {}, env_name = nil)
+    def initialize(hash: {}, env_name: nil)
       @env_name = env_name
       @wrapped = hash.__nanoc_symbolize_keys_recursively
     end
@@ -58,7 +58,7 @@ module Nanoc::Int
         DEFAULT_DATA_SOURCE_CONFIG.merge(ds)
       end
 
-      self.class.new(new_wrapped)
+      self.class.new(hash: new_wrapped)
     end
 
     def with_environment
@@ -70,7 +70,7 @@ module Nanoc::Int
       # Load given environment configuration
       env_config = @wrapped[ENVIRONMENTS_CONFIG_KEY].fetch(env_name.to_sym, {})
 
-      self.class.new(@wrapped, env_name).merge(env_config)
+      self.class.new(hash: @wrapped, env_name: env_name).merge(env_config)
     end
 
     contract C::None => Hash
@@ -108,12 +108,12 @@ module Nanoc::Int
 
     contract C::Or[Hash, self] => self
     def merge(hash)
-      self.class.new(@wrapped.merge(hash.to_h), @env_name)
+      self.class.new(hash: @wrapped.merge(hash.to_h), env_name: @env_name)
     end
 
     contract C::Any => self
     def without(key)
-      self.class.new(@wrapped.reject { |k, _v| k == key }, @env_name)
+      self.class.new(hash: @wrapped.reject { |k, _v| k == key }, env_name: @env_name)
     end
 
     contract C::Any => self
--- a/lib/nanoc/base/entities/configuration.rb
+++ b/lib/nanoc/base/entities/configuration.rb
@@ -39,7 +39,7 @@ module Nanoc::Int
     # Configuration environments property key
     ENVIRONMENTS = :environments
 
-    contract Hash => C::Any
+    contract Hash, C::Maybe[String] => C::Any
     # Creates a new configuration with the given hash.
     #
     # @param [Hash] hash The actual configuration hash
--- a/lib/nanoc/base/entities/configuration.rb
+++ b/lib/nanoc/base/entities/configuration.rb
@@ -37,7 +37,9 @@ module Nanoc::Int
     attr_reader :env_name
 
     # Configuration environments property key
-    ENVIRONMENTS = :environments
+    ENVIRONMENTS_CONFIG_KEY = :environments
+    NANOC_ENV = 'NANOC_ENV'.freeze
+    NANOC_ENV_DEFAULT = 'default'.freeze
 
     contract Hash, C::Maybe[String] => C::Any
     # Creates a new configuration with the given hash.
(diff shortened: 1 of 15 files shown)


barraq:
Is missing:
- handeling `/tmp` dir so that we use the same mechanism as the `output_dir`.
- unit tests

Might be added:
- introduce @env (@site.env) variable to access only active env variable

gpakosz:
Does it take into account the cascading of configuration files (`parent_config_file`)?

gpakosz:
Is `<<: *default` an inheritance mechanism?

barraq:
Good point regarding `parent_config_file`. actually the `config` is overridden after it is loaded, therefore after it include the parent_config_file as well.
Regarding the `<<: *default` it is just some YAML convention: https://en.wikipedia.org/wiki/YAML#References

ddfreyne:
This looks cool! A couple of things that I spotted so far:

* The `#site_from_config` method takes the configuration as an argument, and modifies it. I dislike modifying incoming arguments, because it makes code harder to follow and effects more difficult to trace.

* Whatever’s in `lib/nanoc/base` can’t refer to anything in `lib/nanoc/cli`; the command-line interface knows about `base`, but not vice versa.

barraq:
@ddfreyne I agree with both arguments. I tried to provide another implementation with respect to both points. If you have a nicer approach please could you give me the direction (rough idea on how you would do it).

Regarding the `tmp` dir, I would like to have the same approach as the `output_dir` (output_dir is set as {{output_dir}}/{{env_name}}, if environments is not defined then output_dir remains unchanged) meaning having `tmp_dir` set to {{tmp_dir}}/{{env_name}} if environment is given. What is the best approach to achieve that? Cause as far as I know this `tmp` dir is not a property you can change in configuration, right?

Thanks for your help

lib/nanoc/base/repos/site_loader.rb
```diff
@@ -8,8 +8,9 @@ def new_with_config(hash)
       site_from_config(Nanoc::Int::Configuration.new(hash).with_defaults)
     end
 
-    def new_from_cwd
-      site_from_config(Nanoc::Int::ConfigLoader.new.new_from_cwd)
+    #def new_from_cwd(env = :default)
```
barraq: oups... ➡️ will be removed

lib/nanoc/base/entities/configuration.rb
```diff
@@ -47,6 +47,18 @@ def with_defaults
       self.class.new(new_wrapped)
     end
 
+    def with_environment(env)
```
barraq: Instead of passing `env` as an argument (here and in `new_from_cwd`) would it be correct instead to:
- have in Nanoc::CLI
```
def self.env=(environment)
   ENV.store 'NANOC_ENVIRONMENT', environment
end
```
- have `with_environment` to fetch `NANOC_ENVIRONMENT` in ENV

barraq:
@ddfreyne do you need me to do any additional changes?

ddfreyne:
@barraq I’ll give this a more thorough review soon!

lib/nanoc/base/entities/configuration.rb
```diff
@@ -47,6 +51,21 @@ def with_defaults
       self.class.new(new_wrapped)
     end
 
+    def with_environment(env = nil)
```
ddfreyne: `with_environment` does not appear to be called with an argument except for tests. Make the tests set up the environment, and remove this argument.

lib/nanoc/base/entities/configuration.rb
```diff
@@ -47,6 +51,21 @@ def with_defaults
...
+      return self unless @wrapped.key? :environments
```
ddfreyne: For style consistency, wrap method arguments in parentheses.

lib/nanoc/base/entities/configuration.rb
```diff
@@ -47,6 +51,21 @@ def with_defaults
...
+
+      # Set active environment
+      env ||= ENV.fetch 'NANOC_ENVIRONMENT', :default
```
ddfreyne: 1. Same parentheses remark as above.

2. I suggest `NANOC_ENV`, for consistency with `RAILS_ENV` and `RACK_ENV`

lib/nanoc/base/entities/configuration.rb
```diff
@@ -47,6 +51,21 @@ def with_defaults
...
+
+      # Load given environment configuration
+      env_config = @wrapped[:environments][env.to_sym] || {}
```
ddfreyne: You can use #fetch here: prefer `foo.fetch(a, b)` over `foo[a] || b`

lib/nanoc/base/entities/configuration.rb
```diff
@@ -47,6 +51,21 @@ def with_defaults
...
+
+      # Handle specific properties
+      env_config[:output_dir] ||= File.join(@wrapped[:output_dir].to_s, env.to_s)
```
ddfreyne: The `to_s` calls should not be necessary here.

lib/nanoc/base/repos/checksum_store.rb
```diff
@@ -6,7 +6,7 @@ module Nanoc::Int
   class ChecksumStore < ::Nanoc::Int::Store
     # @param [Nanoc::Int::Site] site
     def initialize(site: nil)
-      super('tmp/checksums', 1)
+      super(File.join('tmp', (site.config.env if site).to_s, 'checksums'), 1)
```
ddfreyne: Can you extract the logic for generating the tmp path into `Nanoc::Int::Store`, as a class method marked `@api private`? Add tests for this extracted method, too. This way, it won’t be duplicated, and it can be tested.

lib/nanoc/base/services/compiler_loader.rb
```diff
@@ -2,10 +2,10 @@ module Nanoc::Int
   # @api private
   class CompilerLoader
     def load(site)
-      rule_memory_store = Nanoc::Int::RuleMemoryStore.new
+      rule_memory_store = Nanoc::Int::RuleMemoryStore.new env: site.config.env
```
ddfreyne: \+ parentheses

lib/nanoc/cli/commands/nanoc.rb
```diff
@@ -10,6 +10,10 @@
   Nanoc::CLI.debug = true
 end
 
+opt :e, :env, 'set environment', argument: :optional do |value|
```
ddfreyne: I don’t think it makes sense for `-e` to have an optional argument, since it defaults to `:default` anyway. Make the argument required, and remove the `|| :default` check below.

lib/nanoc/cli/commands/nanoc.rb
```diff
@@ -10,6 +10,10 @@
...
+  ENV.store 'NANOC_ENVIRONMENT', value || :default
```
ddfreyne: \+ parentheses

spec/nanoc/base/entities/configuration_spec.rb
```diff
@@ -16,4 +15,35 @@
...
   end
+
+  context 'with environments defined' do
+    let(:hash) { { foo: 'bar', environments: { test: { foo: 'test-bar' }, default: { foo: 'default-bar' } } } }
+    let(:config) { described_class.new(hash).with_environment(env) }
+
+    subject { config }
+
+    context 'with existing environment' do
+      let(:env) { :test }
```
ddfreyne: Environment names should be strings.

lib/nanoc/base/entities/configuration.rb
```diff
@@ -31,10 +31,14 @@ class Configuration
       string_pattern_type: 'glob',
     }.freeze
 
+    attr_reader :env
+
     # Creates a new configuration with the given hash.
     #
     # @param [Hash] hash The actual configuration hash
-    def initialize(hash = {})
+    # @param [Symbol, String] env The active environment for this configuration
```
ddfreyne: I don’t think the environment should be allowed to be a symbol.

lib/nanoc/base/entities/configuration.rb
```diff
@@ -31,10 +31,14 @@ class Configuration
       string_pattern_type: 'glob',
     }.freeze
 
+    attr_reader :env
```
ddfreyne: Can you document this? `@return [String, nil]`

lib/nanoc/base/entities/configuration.rb
ddfreyne: I’m not in favour of modifying `output_dir` like this. The output_dir can be specified on a per-environment basis, if necessary, so there’s no need for this logic.

lib/nanoc/base/entities/configuration.rb
ddfreyne: Extract the `:environments` symbol into a constant.

ddfreyne:
Looks good otherwise! Looking forward to using this.

barraq:
Thanks @ddfreyne for this deep review. I fixed all but one:

> I’m not in favour of modifying output_dir like this. The output_dir can be specified on a per-environment basis, if necessary, so there’s no need for this logic.

Actually I did this to prevent having a mandatory argument when using environments. To my understanding if the user does not override the `output_dir` in each defined environment then there will be some weird things happening with the compilation when --prune is activated for instance? Same if the option is not activated since residues from another compilation might remain. I mean that line prevents side effect behavior introduced by environments.

What do you think?

I am fine with removing it :D I just prefer having a second thought before doing it.

One last think; should this PR include the documentation with it? How to do link new feature and the Nanoc.ws documentation?

p.s. sorry for all the parenthesis thingy; I relied on Rubocop.

ddfreyne:
I’m short on time atm, but will revisit as soon as I can.

lib/nanoc/base/entities/configuration.rb
```diff
@@ -51,6 +59,21 @@ def with_defaults
...
+      return self unless @wrapped.key?(ENVIRONMENTS)
+
+      # Set active environment
+      env = @env || ENV.fetch('NANOC_ENV', 'default')
+
+      # Load given environment configuration
+      env_config = @wrapped[ENVIRONMENTS].fetch(env.to_sym, {})
+
+      # Handle specific properties
+      env_config[:output_dir] ||= File.join(@wrapped[:output_dir].to_s, env.to_s)
```
ddfreyne: This mutates `env_config`, which is not good.

lib/nanoc/base/repos/store.rb
```diff
@@ -34,6 +34,12 @@ def initialize(filename, version)
       @version  = version
     end
 
+    # Logic for building tmp path from active environment and store name
+    # @api private
+    def self.tmp_path_for(env, store)
```
ddfreyne: Add a `contract` here as well.

lib/nanoc/base/repos/dependency_store.rb
```diff
@@ -5,8 +5,8 @@ class DependencyStore < ::Nanoc::Int::Store
     attr_reader :objects
 
     # @param [Array<Nanoc::Int::Item, Nanoc::Int::Layout>] objects
-    def initialize(objects)
-      super('tmp/dependencies', 4)
+    def initialize(objects, env: nil)
+      super(Nanoc::Int::Store.tmp_path_for(env, 'dependencies'), 4)
```
ddfreyne: This lacks tests

lib/nanoc/base/repos/rule_memory_store.rb
```diff
@@ -4,8 +4,8 @@ module Nanoc::Int
   #
   # @api private
   class RuleMemoryStore < ::Nanoc::Int::Store
-    def initialize
-      super('tmp/rule_memory', 1)
+    def initialize(env: nil)
+      super(Nanoc::Int::Store.tmp_path_for(env, 'rule_memory'), 1)
```
ddfreyne: This lacks tests

test/base/test_store.rb
```diff
@@ -32,4 +32,16 @@ def test_delete_and_reload_on_error
...
   end
+
+  def test_tmp_path_with_nil_env
+    tmp_path_for_checksum = Nanoc::Int::Store.tmp_path_for(nil, 'checksum')
+    assert_equal('tmp/checksum', tmp_path_for_checksum)
+  end
+
+  def test_tmp_path_with_test_env
+    tmp_path_for_checksum = Nanoc::Int::Store.tmp_path_for('test', 'checksum')
+    tmp_path_for_dependencies = Nanoc::Int::Store.tmp_path_for(:test, 'dependencies')
```
ddfreyne: I’d remove support for symbols, and only use strings instead (there’s no use case for symbols).

lib/nanoc/base/entities/configuration.rb
```diff
@@ -51,6 +59,21 @@ def with_defaults
       self.class.new(new_wrapped)
     end
 
+    def with_environment
+      return self unless @wrapped.key?(ENVIRONMENTS)
+
+      # Set active environment
+      env = @env || ENV.fetch('NANOC_ENV', 'default')
```
ddfreyne: The name `env` (and `@env`) is confusing; the name makes it sound like it’s similar to `ENV`. Maybe `env_name`?

lib/nanoc/base/entities/configuration.rb
```diff
@@ -33,11 +33,19 @@ class Configuration
       string_pattern_type: 'glob',
     }.freeze
 
-    contract Hash => C::Any
+    # @return [String, nil] The active environment for the configuration
+    attr_reader :env
+
+    # Configuration environments property key
+    ENVIRONMENTS = :environments
```
ddfreyne: Rename `ENVIRONMENTS` to `ENVIRONMENTS_CONFIG_KEY`

lib/nanoc/base/entities/configuration.rb
```diff
@@ -33,11 +33,19 @@ class Configuration
       string_pattern_type: 'glob',
     }.freeze
 
-    contract Hash => C::Any
+    # @return [String, nil] The active environment for the configuration
+    attr_reader :env
```
ddfreyne: Also here: rename `env` to `env_name`

lib/nanoc/base/entities/configuration.rb
```diff
@@ -33,11 +33,19 @@ class Configuration
...
+    # Configuration environments property key
+    ENVIRONMENTS = :environments
+
+    contract Hash, C::Maybe[String] => C::Any
     # Creates a new configuration with the given hash.
     #
     # @param [Hash] hash The actual configuration hash
-    def initialize(hash = {})
+    # @param [String, nil] env The active environment for this configuration
+    def initialize(hash = {}, env = nil)
```
ddfreyne: The constructor of this class is becoming not so obvious at this point. Can you use named arguments rather than default values? (e.g. `def initialize(hash: {}, env: nil)`)

lib/nanoc/base/entities/configuration.rb
ddfreyne: Extract `'default'` and `NANOC_ENV` into constants

lib/nanoc/base/entities/configuration.rb
ddfreyne: Actually, remove `output_dir` entirely.

ddfreyne:
> Thanks @ddfreyne for this deep review. I fixed all but one:
>
>> I’m not in favour of modifying output_dir like this. The output_dir can be specified on a per-environment basis, if necessary, so there’s no need for this logic.
> 
> Actually I did this to prevent having a mandatory argument when using environments. To my understanding if the user does not override the `output_dir` in each defined environment then there will be some weird things happening with the compilation when --prune is activated for instance? Same if the option is not activated since residues from another compilation might remain. I mean that line prevents side effect behavior introduced by environments.
> 
> What do you think?
> 
> I am fine with removing it :D I just prefer having a second thought before doing it.

Compiling with one env and pruning with another will cause the pruning to happen as if the site was compiled with the env used for pruning. I think that’s expected behavior.

The current implementation (which sets `output_dir`) is not backwards compatible. Output will be written to `output/default/` by default, rather than `output/`, which is a change in behavior.

Even retaining the current behavior for the default env is problematic, as output will be written to either `output/` or a subdirectory (e.g. `output/devel/` when using the `devel` env), which is even more confusing.

lib/nanoc/base/repos/checksum_store.rb
```diff
@@ -6,7 +6,7 @@ module Nanoc::Int
   class ChecksumStore < ::Nanoc::Int::Store
     # @param [Nanoc::Int::Site] site
     def initialize(site: nil)
-      super('tmp/checksums', 1)
+      super(Nanoc::Int::Store.tmp_path_for((site.config.env if site), 'checksums'), 1)
```
ddfreyne: This lacks tests

lib/nanoc/base/repos/compiled_content_cache.rb
```diff
@@ -4,8 +4,8 @@ module Nanoc::Int
   #
   # @api private
   class CompiledContentCache < ::Nanoc::Int::Store
-    def initialize
-      super('tmp/compiled_content', 2)
+    def initialize(env: nil)
+      super(Nanoc::Int::Store.tmp_path_for(env, 'compiled_content'), 2)
```
ddfreyne: This lacks tests

spec/nanoc/base/repos/config_loader_spec.rb
```diff
@@ -52,6 +52,25 @@
...
+        expect(subject).to be_a(Nanoc::Int::Configuration)
+      end
+
+      it 'has the test environment custom option' do
+        allow(ENV).to receive(:fetch).with('NANOC_ENV', 'default').and_return('test')
+        expect(subject[:foo]).to eq('test-bar')
+      end
+
+      it 'has the default environment custom option' do
+        expect(subject[:foo]).to eq('default-bar')
```
ddfreyne: Add a test for options defined not within the `default` env, but globally (should return `bar`)

lib/nanoc/base/repos/store.rb
ddfreyne: The name `store` is confusing, as it is the _name_ of a store rather than the store itself. Rename to `store_name`.

ddfreyne: When calling this function, the argument names aren’t obvious. Using named arguments will make this nicer.

ddfreyne:
@barraq Is this PR still being worked on? I can take over if you wish.

lib/nanoc/base/repos/dependency_store.rb
barraq: Well I added an extra test here https://github.com/barraq/nanoc/blob/feature/support-environments/test/base/test_store.rb#L41 is that enough? or should I introduce a spec for Store (I didn't find one for it actually)

barraq:
@ddfreyne I went over all fixes you mentioned. I did small fixup commits to ease the review process.
@gpakosz support for environments allows you to override the `@site.config`. You can then use it however you want: e.g. filter (within your `Rules` file) draft items in production environment; display additional things in templates for a given environments; etc.

ddfreyne:
Looks good! Thanks for this feature. Merging!

bburton:
As much as I like the idea of adding environment support, this change makes certain assumptions that will cause problems with certain filters. For instance, the compass filter creates a .sass-cache folder in the project directory where nanoc compile is run. It's quite possible other filters may create temporary folders where files are cached.

As a result, the only safe way to use this feature is to only compile to one environment in a given project folder and not compile different environments in the same project folder. 

Sure it's possible to configure the sass compiler to cache to a specific location but the setup is complicated because the environment setting needs to be passed to the compass config.rb and used to set the [sass :cache_location property](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#cache_location-option). Not only that, the configuration has to be made to config.rb so it's possible to run the compass command by itself outside of nanoc and have it pick up the right environment location.

Compass also supports environments so it would be ideal if the environment setting configured for Nanoc could be passed into the compass config.rb but that's a bit beyond the current topic.

However, aside from the filters nanoc supports out of the box, even if all of them were reviewed and fixed if necessary, there's no guarantee that third-party filters aren't writing or caching files to the project directory.

Even if switching environments to compile into the same folder is not supported, environment support is still very useful when building with one environment configuration such as development in its own folder, then push changes to a Git repository, e.g. github.com, then in a different folder configured as the production environment perform a git pull and recompile.

ddfreyne:
@bburton Thanks for the comment! I’ve collected your thoughts in #937.

PR #862, opened 2016-06-07
Fix generated Atom tags being timezone-dependent

jethrogb:
The [`atom_tag_for`](https://github.com/nanoc/nanoc/blob/master/lib/nanoc/helpers/blogging.rb#L230-L237) function uses a timezone-dependent conversion function `__nanoc_to_iso8601_date` to generate the Atom tag from `created_at`. For example:

```
$ TZ=Europe/Amsterdam ruby -rnanoc -ryaml -e "extend Nanoc::Helpers::Blogging;puts attribute_to_time(YAML::load('2015-03-09 00:00:00')).to_iso8601_date"
2015-03-09
$ TZ=America/Los_Angeles ruby -rnanoc -ryaml -e "extend Nanoc::Helpers::Blogging;puts attribute_to_time(YAML::load('2015-03-09 00:00:00')).to_iso8601_date"
2015-03-08
```

`to_iso8601_time` already uses getutc, so I think `to_iso8601_date` should do the same.

Note: this fix might result in people's tags being changed once when they update nanoc if they were originally running their code in a timezone different from UTC.

### DIFF
--- a/lib/nanoc/extra/core_ext/time.rb
+++ b/lib/nanoc/extra/core_ext/time.rb
@@ -2,7 +2,7 @@
 module Nanoc::Extra::TimeExtensions
   # @return [String] The time in an ISO-8601 date format.
   def __nanoc_to_iso8601_date
-    strftime('%Y-%m-%d')
+    getutc.strftime('%Y-%m-%d')
   end
 
   # @return [String] The time in an ISO-8601 time format.


ddfreyne:
This makes sense! Can you make two small changes:

* Open a PR against the `release-4.2` branch instead, so it ends up in the next bugfix release.
* Add a test for this in `test/extra/core_ext/test_time.rb`.

Happy to help with either!

jethrogb:
Any ideas on how to approach the test? AFAIK there is no platform-independent way to configure the current timezone.

ddfreyne:
I believe you can do `allow(Time).to receive(:zone).and_return('CEST')`. That’d be in the `before` block of a context.

ddfreyne:
I looked into writing specs for this, and it turns out that this change isn’t quite correct—when given a date (rather than a time), the fix fails. Will work on a separate PR that fixes this and adds specs.

jethrogb:
Thanks for picking this up, I've been swamped...

ddfreyne:
`#to_iso8601_date` _deliberately_ does not convert to UTC because the date is supposed to be timezone-agnostic.

For example, if I were to write `2016-10-22`, then I intend that specific day, no matter where I am, and what time it is. Converting this date to UTC first could change that date to `2016-10-21` or `2016-10-23`, which is not what’s intended.

Perhaps `#attribute_to_time` should be updated so that, given a `Date` rather than a `Time`, the generated `Time` object is created from UTC time. Will give that a try!

ddfreyne:
Replaced by #961.

PR #865, opened 2016-06-08
Fix WeakRef::RefError

### LINKED ISSUE #863: WeakRef::RefError: Invalid Reference - probably recycled
Hi,

first thanks for the great and consistent work on nanoc.
Me and my team really appreciate it.

**Problem**

I create the used items inside a `preprocess` rule.
After upgrading to Nanoc 4.2.0 I experience the following error when I access `@item` inside a template:

````
Message:

WeakRef::RefError: Invalid Reference - probably recycled

Compilation stack:

  (empty)
````

**Environment**

```
Nanoc 4.2.0 © 2007-2016 Denis Defreyne.
Running ruby 2.2.4 (2015-12-16) on x64-mingw32 with RubyGems 2.4.5.1.
```

**crash.log**
[crash.txt](https://github.com/nanoc/nanoc/files/305231/crash.txt)

Let me know if I can provide further information  to this topic

Kind Regards
Gregor

ddfreyne:
I assume that checking `#weakref_alive?` is not a strong guarantee that the ref is alive. Perhaps the ref dies after the call, but before it’s accessed?

Needs more rigorous (load) testing.

Fixes #863.

### DIFF
--- a/lib/nanoc/base/memoization.rb
+++ b/lib/nanoc/base/memoization.rb
@@ -15,6 +15,8 @@ module Nanoc::Int
       end
     end
 
+    NONE = Object.new
+
     # Memoizes the method with the given name. The modified method will cache
     # the results of the original method, so that calling a method twice with
     # the same arguments will short-circuit and return the cached results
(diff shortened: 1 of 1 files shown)


ddfreyne:
Load testing shows that this is fine.

PR #875, opened 2016-06-27
Support missing Rouge 1.x options

barraq:
This commit adds support for Rouge 1.x missing options:
- inline_theme (default: nil)
- line_numbers (default: false)
- start_line (default: 1)
- wrap (default: false)

### DIFF
--- a/lib/nanoc/filters/colorize_syntax.rb
+++ b/lib/nanoc/filters/colorize_syntax.rb
@@ -314,6 +314,10 @@ module Nanoc::Filters
 
       formatter_options = {
         css_class: params.fetch(:css_class, 'highlight'),
+        inline_theme: params.fetch(:inline_theme, nil),
+        line_numbers: params.fetch(:line_numbers, false),
+        start_line: params.fetch(:start_line, 1),
+        wrap: params.fetch(:wrap, false),
       }
       formatter = Rouge::Formatters::HTML.new(formatter_options)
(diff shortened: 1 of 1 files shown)


barraq:
Ok @ddfreyne I run the test locally and I got no errors... which version of Rouge is running on Travis?

ddfreyne:
Looks good. Can you also add a test for this?

The failing tests are new Rubocop checks, which I’ve fixed already, so if you rebase onto the `master` branch, these errors will go away.

barraq:
@ddfreyne actually there is a little issue. Since in the Gemfile we do not restrict Rouge to be 1.x it is by default the latest version (being 2.x) and therefore no matter which option you provide they will not be use:

> Rouge::Formatters::HTML.new - will render your code with standard class names for tokens, with no div-wrapping or other bells or whistles.

So actually if you use Rouge 2.x no matter options you provide Rouge does not care. As a result, if I write tests for Rouge 1.x they will brake in Rouge 2.x since options will be discarded.

What about:
- changing the order of commits:
  - first introduce support for Rouge 2.x with Legacy as default
  - then introduce support for Rouge 1.x options and add tests for it
- then later add non legacy support for Rouge 2.x

Is it something ok for you?

ddfreyne:
I’d welcome a PR that supports both Rouge 1.x and 2.x, as long as for Rouge 2.x, the formatter is not hardcoded to be the legacy one. So, plan B but also added non-legacy support.

barraq:
Close PR in favor of https://github.com/nanoc/nanoc/pull/880

PR #880, opened 2016-06-29
Improve support for Rouge 1.x and 2.x

barraq:
This commit improves support for Rouge 1.x and adds support for Rouge 2.x.
- Support for following options are introduced:
  - inline_theme (default: nil)
  - line_numbers (default: false)
  - start_line (default: 1)
  - wrap (default: false)
- Support for formatters (including HTMLLegacy, HTMLInline, HTMLLinewise, HTMLPygments, HTMLTable) is introduced for Rouge 2.x
## Todo:
- [x] Initial (might do better) implementation
- [x] Split colorize filter tests
- [x] Introduce Appraisal gem
- [x] Setup Appraisal for Rouge 1.x and 2.x
- [x] Update Travis to run test for both Rouge 1.x and Rouge 2.x
- [x] Refactor Rouge tests to specs
- [x] Add specific tests for Rouge 2.x regarding HTMLInline
- [x] Add specific tests for Rouge 2.x regarding  HTMLLinewise
- [x] Add specific tests for Rouge 2.x regarding HTMLTable
- [x] Add specific tests for Rouge 2.x regarding HTMLPygments
- [x] Simplified Rouge 2.x implementation (allow any kind of formatters)
- [x] Improve Rake/Travis configuration to only run Rouge specs for Rouge appraisal

## Usage for Rouge 1.x (no change):

```
filter :colorize_syntax,
       :default_colorizer => :rouge,
       :rouge => { line_numbers: true, wrap: true, css_class: 'highlight' }
```
## Usage for Rouge 2.x:

```
# Legacy support
filter :colorize_syntax,
       :default_colorizer => :rouge,
       :rouge => { legacy: true, line_numbers: true, wrap: true, css_class: 'highlight' }

# Custom formatter (here HTML Inline)
filter :colorize_syntax,
       :default_colorizer => :rouge,
       :rouge => { formatters: Rouge::Formatters::HTMLInline.new(MY_THEME) }
```

See Rouge 2.x doc for other formatters


lib/nanoc/filters/colorize_syntax.rb
```diff
@@ -312,10 +312,36 @@ def coderay_postprocess(_language, element)
     def rouge(code, language, params = {})
       require 'rouge'
 
-      formatter_options = {
-        css_class: params.fetch(:css_class, 'highlight'),
-      }
-      formatter = Rouge::Formatters::HTML.new(formatter_options)
+      if Rouge.version < '2' || params.fetch(:legacy, false)
```
barraq: we might change the logic here so that if we have Rouge 2.x and than neither `inline, linewise, pygments, table` is define and than there is arguments given then we fallback to legacy. That would prevent from explicitly specifying legacy.

ddfreyne:
@barraq What is the status of this PR?

barraq:
@ddfreyne I didn't have the time to go over this Travis things to run test for both Rouge 1.x and Rouge 2.x. If you know how to do it quick it would be cool. I can add the missing tests then.

connorshea:
You can use separate gemfiles that use Rouge 1.x vs. 2.x, see [the docs](https://docs.travis-ci.com/user/languages/ruby/#Custom-Bundler-arguments-and-Gemfile-locations).

Personally I'd prefer an approach that just uses the `bundle install --gemfile` approach so you can switch to a different CI system more easily, but that's not really necessary.

barraq:
Thanks @connorshea. Ok @ddfreyne like discussed I am looking at https://github.com/thoughtbot/appraisal

barraq:
@ddfreyne I am almost done here, any comments?

ddfreyne:
This is great work!

I suppose the only thing that’s left is to only run the Rouge specs twice (rather than everything).

ddfreyne:
Question: How are the gemfiles/rouge_1.gemfile and gemfiles/rouge_2.gemfile files generated? Can they be auto-generated, or should they be checked in?

barraq:
Thanks @ddfreyne, regarding `gemfiles/*` they are generated by Appraisal by doing `appraisal install` (see https://github.com/thoughtbot/appraisal#under-the-hood) So they are autogenerated and are based on the initial `Gemfile`.

test/filters/colorize_syntax/test_pigmentize.rb
```diff
@@ -0,0 +1,35 @@
+class Nanoc::Filters::ColorizeSyntax::PigmentizeTest < Nanoc::TestCase
```
ddfreyne: Pygmentize, not pigmentize (the “Py” is from Python!)

Same for the filename.

gemfiles/rouge_1.gemfile
```diff
@@ -0,0 +1,62 @@
+# This file was generated by Appraisal
```
ddfreyne: Are these gemfiles still necessary, given that `appraisal install` is executed already?

ddfreyne:
Looks great! Just some minor comments.

gemfiles/rouge_1.gemfile
barraq: True, actually those gemfiles are only needed when we want to run all the tests. Since in our case we only wants to run `focus` spec we can remove them.

barraq:
@ddfreyne ready to merge?

ddfreyne:
Done. Thank you for your effort!

whitequark:
\o/

axilleas:
Awesome! Does that mean 4.3.8 is about to be published? :smile:

ddfreyne:
Soon!

ddfreyne:
Done!

connorshea:
@ddfreyne @barraq thank you! :)

PR #887, opened 2016-07-01
Use Hamster::Set in checksummer for speed

### LINKED ISSUE #864: Major performance regression 3.7.4 -> 3.7.5 (18 seconds -> 1342 seconds)
On 3.7.1, 3.7.2, 3.7.4 our site generates in under 20 seconds:

`Site compiled in 17.92s.`

On 3.7.5:

`Site compiled in 1341.51s.`

(Most of the delay is before the first line of output after `Compiling site...` is output - I didn't watch for long enough to see it complete but it didn't output anything for at least 5 minutes.)

The only difference otherwise in the output is that 3.7.5 outputs a warning about halfway through logging the outputs; but that's much later than the delay seems to occur.

`kramdown warning: No link definition for link ID 'due to change in future' found on line 9`

How can I help debug this further?

ddfreyne:
Using Ruby’s `Set` in an immutable way is slow. `Hamster::Set`, from [Hamster](https://github.com/hamstergem/hamster), is significantly faster.

Example: 1138s &rarr; 50s

Fixes #864.

### DIFF
--- a/lib/nanoc/base/checksummer.rb
+++ b/lib/nanoc/base/checksummer.rb
@@ -1,3 +1,5 @@
+require 'hamster'
+
 module Nanoc::Int
   # Creates checksums for given objects.
   #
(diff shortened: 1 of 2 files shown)


lib/nanoc/base/checksummer.rb
```diff
@@ -46,14 +48,14 @@ def calc(obj, digest_class = CompactDigest)
...
-      def update(obj, digest, visited = Set.new)
+      def update(obj, digest, visited = Hamster::Set.new)
         digest.update(obj.class.to_s)
 
         if visited.include?(obj)
           digest.update('<recur>')
         else
           digest.update('<')
-          behavior_for(obj).update(obj, digest) { |o| update(o, digest, visited + [obj]) }
+          behavior_for(obj).update(obj, digest) { |o| update(o, digest, visited.add(obj)) }
```
ddfreyne: I could’ve left it that way, but `#add` is a nicer API, and it’s likely faster as well.

PR #889, opened 2016-07-01
Fix specs on Ruby 2.4

ddfreyne:
Ruby 2.4 no longer has `Fixnum`, but has `Integer` instead.

Sadly, the tests are broken on Ruby 2.4 due to a non-updated YAJL:

```
yajl_ext.c:881:22: error: ‘rb_cFixnum’ undeclared (first use in this function)
```


ddfreyne:
Merged this into release-4.3.x. The tests still fail on Ruby 2.4, but will likely start working as soon as YAJL adds Ruby 2.4 support.

PR #909, opened 2016-07-12
Provide overrideable attr methods in blogging helper

ddfreyne:
This adds the methods

* `#article?`
* `#article_created_at`
* `#article_updated_at`

… which are then used in other methods, such as `#articles` and `#sorted_articles`, to find items. This approach allows the mechanism of finding, ordering, … to be customised by overriding these methods.

This is particularly useful because it removes the need to set `kind` to `article` in the preprocessor; as an alternative, `#article?` can be overridden to do what would otherwise be done in the preprocessor.

## Alternative ideas

The idea of providing functions that can be overridden is not something that is present elsewhere in Nanoc, and I’m not such a big fan of this approach. As an alternative, there could be a class that contains/describes the mechanism for finding/ordering items. For example:

```ruby
class BloggingBehavior
  def article?(_item)
    raise NotImplementedError
  end

  # … other functions here …
end

class DefaultBloggingBehavior
  def article?(_item)
    item[:kind] == 'article'
  end

  # … other functions here …
end

class StoneshipSiteBloggingBehavior
  def article?(item)
    item.identifier =~ '/articles/*'
  end

  # … other functions here …
end

blogging_behavior(StoneshipSiteBloggingBehavior)
```

This behavior class could also implement memoized versions of methods (e.g. `#memoized_article?`) which would work transparently in subclasses.

### DIFF
--- a/lib/nanoc/helpers/blogging.rb
+++ b/lib/nanoc/helpers/blogging.rb
@@ -1,9 +1,26 @@
 module Nanoc::Helpers
   # @see http://nanoc.ws/doc/reference/helpers/#blogging
   module Blogging
+    # @param item
+    #
+    # @return [Boolean]
+    def article?(item)
+      item[:kind] == 'article'
+    end
+
+    # TODO: document
+    def article_created_at(item)
+      attribute_to_time(item[:created_at])
+    end
+
+    # TODO: document
+    def article_updated_at(item)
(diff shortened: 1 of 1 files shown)


ddfreyne:
Closing this for now, as it needs more thought and discussion.

PR #910, opened 2016-07-13
Discover external commands through Bundler

whitequark:
Other gems may want to add commands to the nanoc CLI. To do this
they need to run Nanoc::CLI.add_after_setup_proc. Currently, there
is no way to hook into the CLI. This commit adds such a way by
utilizing Bundler. E.g. the gem guard-nanoc could introduce new
commands when added to the Gemfile in this way:

    gem 'nanoc'
    group :nanoc do
      gem 'guard-nanoc'
    end

Presence of any gems in the group :nanoc will cause the library with
the same name as the gem to be required by Bundler during
the initialization of the CLI.

A special group :nanoc is used for two reasons:
  * nanoc may not be the only or primary user of the Gemfile,
    and requiring many unrelated gems will slow down every invocation
    of nanoc with no workaround;
  * existing Gemfiles may include specifications for gems that cannot
    be successfully required, so using the group :default would
    break such existing nanoc installations.

### DIFF
--- a/lib/nanoc/cli.rb
+++ b/lib/nanoc/cli.rb
@@ -110,6 +110,11 @@ module Nanoc::CLI
       cmd = load_command_at(cmd_filename)
       add_command(cmd)
     end
+
+    if defined?(Bundler)
+      # Discover external commands through Bundler
+      Bundler.require(:nanoc)
+    end
   end
 
   # Loads site-specific commands.


ddfreyne:
Looks good!

The group name `nanoc` doesn’t convey the idea that these are gems that will be automatically required. Perhaps a group names`autorequire`?

Additionally, since Bundler is not a runtime requirement of Nanoc, `Bundler.require(:nanoc)` can potentially fail.

whitequark:
> The group name nanoc doesn’t convey the idea that these are gems that will be automatically required. Perhaps a group named autorequire?

I'm not sure if such change would make a lot of sense. First, there is no guarantee that the entire Gemfile would not be autorequired; indeed, whether or not to require gems is a choice of an application using the Gemfile. Second, autorequiring groups of gems is, as far as I'm aware, literally the purpose of the Bundler `group` feature (e.g. see its [docs](http://bundler.io/v1.11/groups.html)), so that should not come as a surprise.

> Additionally, since Bundler is not a runtime requirement of Nanoc, Bundler.require(:nanoc) can potentially fail.

I've wrapped that in `if defined?(Bundler)`.

ddfreyne:
>> The group name nanoc doesn’t convey the idea that these are gems that will be automatically required. Perhaps a group named autorequire?
>
> I'm not sure if such change would make a lot of sense. First, there is no guarantee that the entire Gemfile would not be autorequired; indeed, whether or not to require gems is a choice of an application using the Gemfile. Second, autorequiring groups of gems is, as far as I'm aware, literally the purpose of the Bundler `group` feature (e.g. see its [docs](http://bundler.io/v1.11/groups.html)), so that should not come as a surprise.

Thinking about it again, the group name `nanoc` makes sense (as in, these are the gems that nanoc will load by default).

lib/nanoc/cli.rb
```diff
@@ -110,6 +110,11 @@ def self.setup_commands
       cmd = load_command_at(cmd_filename)
       add_command(cmd)
     end
+
+    if defined?(Bundler)
+      # Discover external commands through Bundler
+      Bundler.require(:nanoc)
```
ddfreyne: What happens if no `nanoc` group exists? What if it is empty?

A change without tests puts me off, but at the same time, I have no idea how to test this in an automated fashion.

whitequark: Nothing. That's how this Bundler API works.

whitequark:
Thanks!

PR #911, opened 2016-07-18
Add resource uris to ilinks check

werthen:
Right now the ilinks check only checks for internal links in `a` and `img` tags. Just as an example: I recently refactored the locations of my assets, and wanted to locate all the places in my code where these references were broken, but `nanoc check ilinks` stated everything was a-ok.

This PR makes `nanoc check ilinks` also check for resource links (`style`, `script`, ...)


ddfreyne:
This looks great! Can you ensure that the tests also test the new functionality?

ddfreyne:
Also, as I see this as either an enhancement or bug fix, could you base the branch of release-4.2.x, so that this ends up in the next 4.2.x release?

Issue #900, opened 2016-07-03
On Windows we either get UTF-8 or colored output


agross:
Hi,

I hope you are okay with the amount of issues I recently report. (Let me know if you're not.)

I'm running Windows and Linux in parallel. It seems that Nanoc on Windows cannot output UTF-8 and colors at the same time.

There are several shells/terminals available on my machine and there's also Nanoc's check for the `win32console` gem:

* default Windows 10 terminal, e.g. `conhost.exe` + `cmd.exe`
* ConEmu running `zsh`

| Shell/Terminal | `ENV['LANG'] =~ /UTF-8/i` | `win32console` available | colors enabled | colors working | UTF-8 enabled | UTF-8 working | As expected?  |
| :------------- | :-----------------------: | :----------------------: | :------------: | :------------: | :-----------: | :-----------: | :-----------: |
| cmd.exe        | ✘                         | ✘                        | ✘              | ✘              | ✘             | ✘             | ✔             |
| cmd.exe        | ✘                         | ✔                        | ✔              | ✔              | ✘             | ✘             | ✔             |
| cmd.exe        | ✔                         | ✘                        | ✘              | ✘              | ✔             | ✔             | ✔             |
| cmd.exe        | ✔                         | ✔                        | ✔              | ✔              | ✔             | ✘             | ✘             |
| ConEmu         | ✘                         | ✘                        | ✘              | ✘              | ✘             | ✘             | ✔             |
| ConEmu         | ✘                         | ✔                        | ✔              | ✔              | ✘             | ✘             | ✔             |
| ConEmu         | ✔                         | ✘                        | ✘              | ✘              | ✔             | ✔             | ✔             |
| ConEmu         | ✔                         | ✔                        | ✔              | ✔              | ✔             | ✘             | ✘             |

The most important part is the last column where the rows have ✘es. It seems as if `win32console` breaks UTF-8 output.

Interestingly, with `Nanoc::CLI.setup_cleaning_streams` commented, coloring + UTF-8 always works without requiring `win32console` to be available.

I think it's important to note that Microsoft did some work regarding the console in Windows 10. For example, on Windows 7 Ruby cannot `print 'Loading site… '` correctly (it prints `Loading site.`) whereas on Windows 10 this is not an issue.

ddfreyne:
Thanks for the bug reports, first of all—they’re most definitely useful, as Nanoc keeps getting better.

[...]

PR #918, opened 2016-07-24
Remove win32console require

ddfreyne:
win32console (https://github.com/luislavena/win32console) is deprecated
in favour of ANSICON (https://github.com/adoxa/ansicon).

Hopefully makes #900 less of an issue.


ddfreyne:
CC @agross

PR #926, opened 2016-08-03
Allow turning on line numbers for the Rouge formatter

whitequark:
... similarly to what CodeRay allows.

### DIFF
--- a/lib/nanoc/filters/colorize_syntax.rb
+++ b/lib/nanoc/filters/colorize_syntax.rb
@@ -316,6 +316,9 @@ module Nanoc::Filters
         css_class: params.fetch(:css_class, 'highlight'),
       }
       formatter = Rouge::Formatters::HTML.new(formatter_options)
+      if params[:line_numbers]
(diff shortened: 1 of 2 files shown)


ddfreyne:
Looks interesting. Do you think it makes more sense to allow passing a custom formatter instance (`formatter: Rouge::Formatters::Whatever.new(…)`)?

FYI: If you rebase on top of `master`, the build failures will go away.

whitequark:
I think it's convenient to not have to care about Rouge's API for anyone using nanoc for basic tasks.

whitequark:
Rebased.

barraq:
@whitequark this is connected to https://github.com/nanoc/nanoc/pull/880 did you try out?

whitequark:
@barraq I've missed that PR. I've looked at it and yeah, that should work for me as well.

ddfreyne:
I think my preference goes out for #880 rather than this PR, as it’s more generic and solves the same problem.

whitequark:
Well... this PR is done and #880 is not. But alright.

PR #929, opened 2016-08-14
Lock ruby_dep to 1.3.x on Ruby 2.1.x

ddfreyne:
Recent version  of `ruby_dep` require Ruby 2.2+. Since Nanoc still runs tests on Ruby 2.1, let’s lock ruby_dep to 1.3.x on Ruby 2.1.x.


ddfreyne:
Ruby 2.1 is no longer supported; my mistake.

PR #935, opened 2016-08-22
Collect anonymous usage information

ddfreyne:
This PR collects anonymous usage information, so that it can be sent to nanoc.ws for collection and aggregation. This information will be useful in the long run to determine how Nanoc is used in the real world.

Also see [RFC: metrics collection](https://github.com/nanoc/rfcs/pull/6).

### Bits to collect

Generic event information:

* Date
* Ruby version (`RbConfig::CONFIG` — `'MAJOR'`, `'MINOR'`)
* Nanoc version (`Nanoc::VERSION`)
* OS (`RbConfig::CONFIG['arch']` — also see [Rubygems’ `plaform.rb`](https://github.com/rubygems/rubygems/blob/v2.6.3/lib/rubygems/platform.rb#L19-L112))

Events:

* Event: Nanoc started
* Event: Site about to be compiled
  * Configuration details
    * `string_pattern_type`
    * `identifier_type`
    * `enable_output_diff`
    * `auto_prune`
  * Site details
    * Number of items, as base-10 logarithm
    * Number of layouts, as base-10 logarithm
* Event: Site compiled
  * Duration
  * Number of outdated items

### To do

* [ ] Determine and implement bits to collect
* [ ] Determine when to collect (every command invocation? only compile?)
* [ ] Only ask when in interactive mode (using `Nanoc::CLI.tty?`)
* [ ] Send to nanoc.ws
* [ ] Add tests
* [ ] Add documentation
* [ ] Add opt-in question on startup


gpakosz:
Please make it opt-in up front. We happen to build nanoc docs on build nodes that are isolated from the mighty internet

ddfreyne:
@gpakosz Good point. Nanoc can detect whether it runs in interactive mode (when using a tty), so it can avoid asking when run in an automated fashion.

ddfreyne:
The design of this will live in the [metrics collection RFC](https://github.com/nanoc/rfcs/pull/6), as is too large of an effort to implement without designing it first.

PR #946, opened 2016-09-26
Prune before compiling (fixes #942)

### LINKED ISSUE #942: Compilation fails when creating a folder when file exists in output folder
### Summary

File does not get deleted in output folder prior to creating new folder.

### Steps to reproduce

1. make file without extension
2. make a route to create a folder with the file name of step 1

### Expected behavior

Delete the file, create the folder.

### Actual behavior

Crashes

### Crash log

https://gist.github.com/myncke/e2aeed489eda1d3131e08385724919df

ddfreyne:
Prune before compiling, to delete any files/directories that could interfere with new files/directories being generated. Fixes #942.

### DIFF
--- a/lib/nanoc/cli/commands/compile.rb
+++ b/lib/nanoc/cli/commands/compile.rb
@@ -403,8 +403,8 @@ module Nanoc::CLI::Commands
 
       puts 'Compiling site…'
       run_listeners_while do
-        site.compile
         prune
+        site.compile
       end
 
       time_after = Time.now
--- a/spec/nanoc/regressions/gh_942_spec.rb
+++ b/spec/nanoc/regressions/gh_942_spec.rb
@@ -1,0 +1,21 @@
+describe 'GH-942', site: true, stdio: true do
+  before do
+    File.write('content/foo.md', 'Foo!')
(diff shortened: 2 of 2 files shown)


agross:
Unfortunately, this fails if `site.config[:output_dir]` does not exist at the time the pruner runs.

(My output directory is `build/bin/web/bin`)

```ruby
>> Find.find('foo/bar/baz/') do |f| puts f; end
Errno::ENOENT: No such file or directory
        from C:/Ruby/lib/ruby/2.3.0/find.rb:43:in `block in find'
        from C:/Ruby/lib/ruby/2.3.0/find.rb:43:in `collect!'
        from C:/Ruby/lib/ruby/2.3.0/find.rb:43:in `find'

>> Find.find('foo/') do |f| puts f; end
Errno::ENOENT: No such file or directory
        from C:/Ruby/lib/ruby/2.3.0/find.rb:43:in `block in find'
        from C:/Ruby/lib/ruby/2.3.0/find.rb:43:in `collect!'
        from C:/Ruby/lib/ruby/2.3.0/find.rb:43:in `find'
        from (irb):11
        from C:/Ruby/bin/irb:11:in `<main>'
```

PR #959, opened 2016-10-22
Add CLI query mechanism

ddfreyne:
Allow executing commands without spawning a shell.

### To do

* [x] Tests
* [ ] Documentation
* [x] Hide behind feature flag
* [ ] Move into a separate command

### Detailed description

`shell` currently opens a shell where commands can be executed. This changes the `shell` command to _also_ allow executing arbitrary commands.

For example, given a site with a `/about.md` file:

```bash
$ nanoc shell "@items['/about.*'].identifier.ext"
md
$
```

**NOTE:** Reusing the `shell` command for this purpose is not nice. This should be extracted into a different command (e.g. `query`).

### DIFF
--- a/lib/nanoc/base/feature.rb
+++ b/lib/nanoc/base/feature.rb
@@ -63,22 +63,6 @@ module Nanoc
     end
 
     # @api private
-    def self.enable(feature_name)
-      raise ArgumentError, 'no block given' unless block_given?
-
-      if enabled?(feature_name)
-        yield
-      else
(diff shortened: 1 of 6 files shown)


ddfreyne:
Question: what is the expected output of this command?

* Human-readable
* JSON
* Ruby `#inspect` stuff
* …

ddfreyne:
Closing for now, as there’s no good use case and open questions remain.

PR #961, opened 2016-10-22
Fix timezone issues in #attribute_to_time/#to_iso8601_date

ddfreyne:
## Summary

* Make `#attribute_to_time` convert `Date` to UTC (rather than local) `Time`
* Make `#to_iso8601_date` return UTC representation

## Detailed description

`#to_iso8601_date` currently deliberately does not convert to UTC, because the date is supposed to be timezone-agnostic.

For example, if I were to write `2016-10-22`, then I intend that specific day, no matter where I am, and what time it is. Converting this date to UTC first could change that date to `2016-10-21` or `2016-10-23`, which is not what’s intended.

Unfortunately, this does not hold up when I write a `Time` (e.g. `2016-10-22 10:15:00`) rather than a `Date` and then run it through `#to_iso8601_date`.

In order to make both cases work,  this PR converts `Date` to UTC `Time`s rather than local `Time`s, and then makes `#to_iso8601_date` convert to UTC.

This PR also uncovered an interesting case in the XML sitemap helper (arguably a bug), which is now also fixed.

## See also

* #862 (original fix for this issue)

### DIFF
--- a/lib/nanoc/extra/core_ext/time.rb
+++ b/lib/nanoc/extra/core_ext/time.rb
@@ -2,7 +2,7 @@
 module Nanoc::Extra::TimeExtensions
   # @return [String] The time in an ISO-8601 date format.
   def __nanoc_to_iso8601_date
-    strftime('%Y-%m-%d')
+    getutc.strftime('%Y-%m-%d')
   end
 
   # @return [String] The time in an ISO-8601 time format.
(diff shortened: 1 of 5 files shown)


ddfreyne:
CC @jethrogb — this is the continuation of your PR that you originally opened!

Issue #675, opened 2015-07-11
Option to disable/filter kramdown warnings


ddfreyne:
kramdown spits out warnings that are not always very useful. For instance:

```
kramdown warning: No link definition for link ID 'git' found on line 4 
```

It would be useful to have an option to either disable warnings entirely, or filter them.

gpakosz:
I started implementing warning filtering.

My idea was to store regexps inside `@config[:kramdown][warning_filters]` as an array. Later, I would filter with `filter :kramdown, @config[:kramdown] || {}`. Then inside the `:kramdown` filter implementation, I would use the `warning_filters` array.

At that point, I found it difficult to reliably store regexps inside `nanoc.yaml`. Should we use `!ruby/regexp` or store them as plain strings? If stored as plain strings, are there corner cases when trying to converting them to real regexps in order to filter Kramdown's `@warnings` array?

ddfreyne:
> Should we use `!ruby/regexp` or store them as plain strings?

I think I prefer strings.

> If stored as plain strings, are there corner cases when trying to converting them to real regexps in order to filter Kramdown's `@warnings` array?

Not that I know of… `Regexp.new(string)` should work.

PR #967, opened 2016-10-23
Kramdown warnings

gpakosz:
### Summary

This PR does two things in two commits:
  - tell which item and rep produces a kramdown warning
  - implements warning filtering (see #675) as kramdown itself as no warning switches

### To do

The `:warning_filters` array should be documented on https://nanoc.ws as part of the filters reference documentation.

### DIFF
--- a/lib/nanoc/filters/kramdown.rb
+++ b/lib/nanoc/filters/kramdown.rb
@@ -12,8 +12,11 @@ module Nanoc::Filters
     def run(content, params = {})
       document = ::Kramdown::Document.new(content, params)
 
-      document.warnings.each do |warning|
-        $stderr.puts "kramdown warning: #{warning}"
+      if document.warnings.length != 0
+        $stderr.puts "kramdown warning(s) for #{@item_rep.inspect}"
+        document.warnings.each do |warning|
+          $stderr.puts "  #{warning}"
+        end
       end
 
       document.to_html
--- a/lib/nanoc/filters/kramdown.rb
+++ b/lib/nanoc/filters/kramdown.rb
@@ -10,11 +10,20 @@ module Nanoc::Filters
     #
     # @return [String] The filtered content
     def run(content, params = {})
+      params = params.dup
+      warning_filters = params.delete(:warning_filters)
       document = ::Kramdown::Document.new(content, params)
 
-      if document.warnings.length != 0
+      if warning_filters
+        r = Regexp.union(warning_filters)
+        warnings = document.warnings.reject { |warning| r =~ warning }
+      else
+        warnings = document.warnings
+      end
+
+      if warnings.any?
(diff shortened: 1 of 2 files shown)


test/filters/test_kramdown.rb
```diff
@@ -12,15 +12,41 @@ def test_filter
...
-      filter = ::Nanoc::Filters::Kramdown.new
+      filter = ::Nanoc::Filters::Kramdown.new(item: item_view, item_rep: item_rep_view)
 
       # Run filter
       io = capturing_stdio do
         filter.setup_and_run('{:foo}this is bogus')
       end
       assert_empty io[:stdout]
-      assert_equal "kramdown warning: Found span IAL after text - ignoring it\n", io[:stderr]
+      assert_equal "kramdown warning(s) for item, rep: #{item.identifier}, #{item_rep.name}\n  Found span IAL after text - ignoring it\n", io[:stderr]
```
ddfreyne: I’d replace

```
"… item, rep: #{item.identifier}, #{item_rep.name}"
```

with

```
"… #{item_rep.inspect}"
```

ddfreyne:
Looks good! There’s a couple of style issues that Travis CI complains about still.

lib/nanoc/filters/kramdown.rb
```diff
@@ -10,10 +10,22 @@ class Kramdown < Nanoc::Filter
...
     def run(content, params = {})
+      params = params.dup
+      warning_filters = params.delete(:warning_filters)
       document = ::Kramdown::Document.new(content, params)
 
-      document.warnings.each do |warning|
-        $stderr.puts "kramdown warning: #{warning}"
+      if (warning_filters)
+        r = Regexp.union(warning_filters)
+        warnings = document.warnings.select { |warning| !(r =~ warning) }
```
ddfreyne: Replace

```
!(r =~ warning)
```

with

```
r !~ warning
```

Or alternatively, replace

```
select { |warning| !(r =~ warning) }
```

with

```
reject { |warning| r =~ warning }
```

ddfreyne:
Does it make sense to provide the warning filters via the configuration file, rather than arguments to the `filter` call? That’d make the `dup` dance unnecessary, and I guess it’d avoid having to copy the filter warnings if you call the `:kramdown` filter more than once…

gpakosz:
I took your comments into account.

Well in my setup, I was already doing e.g.
```
# erb filter options
erb:
  trim_mode: '<>'
```

and then when filtering 
```
filter :erb, @config[:erb] || {}
```

Are there other filters reading straight from `@config` ?

connorshea:
This is awesome, I've wanted this myself. Thanks @gpakosz!

ddfreyne:
I’m OK with leaving it as a filter argument, as it’s more flexible, and still leaves open the option for using the configuration in the way you described.

Is this still WIP or is it ready to go?

Issue #972, opened 2016-11-01
Getting "Killed" on Ubuntu 16.04.1 LTS


RubenVerborgh:
### Summary

Running nanoc 4.3.7 (as well as 4.3.2) stops at random moments, emitting "Killed".

### Steps to reproduce

1. Use Ubuntu 16.04.1 LTS
2. Load my site (code not public but available on request)
3. Execute "bundle exec nanoc co"

### Expected behavior

The site compiles.
If an error occurs, I get a stack trace.

### Actual behavior

At random moments in the compilation process, I'm receiving a "Killed" message.

### Details

Sometimes compilation succeeds. Used to work perfectly on other Linuxes on OSX.

### Crash log

None.

RubenVerborgh:
I would have tried to find a MWE, but that's hard without a stack trace.

ddfreyne:
Which version of Ruby are you running?

`lib/nanoc/cli/commands/compile.rb` does some icky tricks with the GC in order to speed up compilation. Usually the icky trick does a remarkable job, but perhaps it’s blowing up in your case. Can you change `GCController.enable_for?` to return false, and see whether the problem persists?

RubenVerborgh:
Running Ruby 2.3.1p112.

Changed it to `true`, ran it 5 times successively, the issue did not occur anymore. (Not a guarantee, however, as the issue does not always occur in general.)

ddfreyne:
Fix in #975.

RubenVerborgh:
Thanks! And interesting about the speed increase 😄

RubenVerborgh:
@ddfreyne Still getting killed occasionally. Could we reopen this?

ddfreyne:
Okay :(

ddfreyne:
Do you have large files that you load in memory in your site? Memory usage is the only thing that I can think of right now…

RubenVerborgh:
I don't really have large files, and certainly not loading them in memory. However, it might be related to memory somehow: when I have `htop` open, CPU initially rises to 26.5% (of 1GB RAM), and then briefly hits 33.3% or so before crashing.

Can I do anything to make more relevant debugging output visible?

ddfreyne:
I’m regularly getting “killed” errors on Travis CI now, so I’ll investigate and fix those, hoping that it’ll also benefit your case.

RubenVerborgh:
[...]

PR #975, opened 2016-11-04
Remove GCController

ddfreyne:
Stop playing around with the GC manually.

### Detailed description

Previous versions of Nanoc (3.x) had the problem of occasionally running very slowly due to spending large amounts of time in the garbage collector (79% for one site).

As a temporary workaround until either Nanoc and/or Ruby improved memory handling, I added a `GCController` which disables and enables the GC at certain times in order to eliminate needless GC activity. I knew from the start that this was a hack and would cause problems, but luckily very few people reported having problems with it—until now, I guess (#972).

Ruby 2.2 and up have generational GC, which greatly reduces the demand on the garbage collector. This should eliminate the need for this GC hack.

### Related issues

* #972 — “Getting "Killed" on Ubuntu 16.04.1 LTS”

### DIFF
--- a/lib/nanoc/cli/commands/compile.rb
+++ b/lib/nanoc/cli/commands/compile.rb
@@ -257,36 +257,6 @@ module Nanoc::CLI::Commands
       end
     end
 
-    # Controls garbage collection so that it only occurs once every 20 items
-    class GCController < Listener
-      # @see Listener#enable_for?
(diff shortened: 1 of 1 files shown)


ddfreyne:
Interesting: removing the GC speed hack leads to a 10%-20% speed improvement for nanoc.ws here.

PR #984, opened 2016-11-15
Use hard dependencies to pick reps to recompile

### LINKED ISSUE #977: ItemRepSelector should prefer unprocessed items
The `ItemRepSelector` does not pick items to recompile in the most efficient order.

Theory: If an item is known to have a dependency (because attempting to compiling it raised an `UnmetDependency`), it is more likely to have dependencies than an item that has not yet been processed yet.

For example, a blog archive page, a tag page, … will collect the compiled content of many items, and thus has many dependencies. It makes sense for Nanoc to attempt compiling all other items before re-attempting compiling the item that generated the dependency.

CC @gpakosz

ddfreyne:
⚠️ **EXPERIMENTAL** ⚠️

1. Add support for hard dependencies to the dependency tracking mechanism. A hard dependency indicates that the dependent item needs to be compiled in order for the dependency to be fulfilled. An `UnmetDependency` error will always result in such a hard dependency.

2. Make the compiler not just compile all outdated items, but also all item reps with a hard dependency from any of these. This fixes #977.

Further work could include making the item rep selector use this hard dependency graph as a hint for which item rep to compile next. This is out of scope for this PR, however.

### DIFF
--- a/lib/nanoc/base/compilation/compiler.rb
+++ b/lib/nanoc/base/compilation/compiler.rb
@@ -207,11 +207,15 @@ module Nanoc::Int
 
       # Find item reps to compile
       outdated_reps = @reps.select { |r| outdatedness_checker.outdated?(r) }
-      dependent_reps = outdated_reps.flat_map { |r| @dependency_store.objects_needed_for_compiled_content_of(r) }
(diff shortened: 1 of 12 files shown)


ddfreyne:
Closing in favour of #985.

PR #985, opened 2016-11-16
Use fibers to minimize redundant work

ddfreyne:
## Summary

Perform compilation in a resumable fiber.

## Motivation

The compilation of an item rep can currently fail due to an unmet dependency (a `Nanoc::Int::Errors::UnmetDependency` exception will be raised). This is suboptimal for two reasons:

* All work done on compiling this item representation is lost, and will have to be started over.

* In a worst-case scenario, an item representation that has many hard dependencies on other other item representations might be compiled in a way that causes the item representation to be compiled over and over again, and failing due to unmet dependencies over and over again. This could drive the compilation time up by orders of magnitude.

The approach in this PR eliminates used work. When an unmet dependency occurs, compilation will continue at the point where the unmet dependency happens.

## Detailed design

The main change is in `lib/nanoc/base/compilation/compiler.rb`, whose diff is best viewed with [?w=1](https://github.com/nanoc/nanoc/pull/985/files). The fiber itself does the work, and the fiber runloop does the notifications. Fibers are retained between `#compile_rep` calls, in order to make them resumable.

The `raise` in `raise Nanoc::Int::Errors::UnmetDependency` has been replaced with `Fiber.yield`. This is the mechanism that allows compilation to be _suspended_ rather than aborted. This also makes the icky `ItemRep#forget_progress` unnecessary, and has been removed.

The `compile` command has gained a few tests for the file action printer and the timing recorded. These have never been tested before, so it’s good to finally have useful tests. I’ve also introduced Timecop, which makes time-based tests more reliable _and_ faster to run (because there is no need for `#sleep`). The `compile` command also had to be modified to take the `:compilation_suspended` notification into account.

The `#compile_rep` method now has unit tests, which unfortunately are somewhat ugly as the method is private, and has a bunch of dependencies which need to be injected. This might mean that this code could/should be extracted; perhaps a SingleRepCompiler might be useful. Such a refactoring is out of scope, though.

## To do

* [x] Save fiber and resume it
* [x] Retry after yielding fiber
* [x] Record timing properly
* [x] Fix compiler stack

Tests:

* [x] `#compile_rep` stack
* [x] `#compile_rep` dependency tracking
* [x] `#compile_rep` notification emission
* [x] `compile` command timing for file actions
* [x] `compile` command timing for filter aggregates

## Related issues

* #977

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,7 @@ group :devel do
   gem 'rspec'
   gem 'rubocop'
   gem 'simplecov', require: false
+  gem 'timecop'
   gem 'vcr'
   gem 'webmock'
   gem 'yard'
--- a/lib/nanoc.rb
+++ b/lib/nanoc.rb
@@ -27,6 +27,7 @@ require 'ref'
 # Load general requirements
 require 'digest'
 require 'enumerator'
+require 'fiber'
 require 'fileutils'
 require 'forwardable'
 require 'pathname'
--- a/lib/nanoc/base/compilation/compiler.rb
+++ b/lib/nanoc/base/compilation/compiler.rb
@@ -227,30 +227,44 @@ module Nanoc::Int
     #
     # @return [void]
     def compile_rep(rep)
-      dependency_tracker = Nanoc::Int::DependencyTracker.new(@dependency_store)
+      @fibers ||= {}
+      @fibers[rep] ||=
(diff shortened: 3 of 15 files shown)


ddfreyne:
This change is not compatible with the presence of the compilation stack. It might be worth removing this stack.

PR #991, opened 2016-11-19
More granular dependencies

ddfreyne:
⚠️ **EXPERIMENTAL** ⚠️

Check outdatedness and record dependencies based on four types of data:

* raw content
* attributes
* compiled content
* path

### Detailed description

* [x] Let outdatedness checker return details on what property is outdated.

* [x] Let outdatedness checker use the dependency properties to determine whether the dependency is causing outdatedness or not.

* [ ] Let dependency tracker/store handle these more granular dependencies.

### To do

Many things. Will fill up this section as I progress.


ddfreyne:
Closing in favour of #995 + future work.

PR #993, opened 2016-11-19
Enable environments feature

ddfreyne:
This removes the feature flag around the environments feature and makes it officially supported. Best viewed with [?w=1](https://github.com/nanoc/nanoc/pull/993/files?w=1).

This requires a 4.4 release.


Fjan:
👍 @ddfreyne Awesome! I'm not sure this gets said enough

Issue #999, opened 2016-11-23
Less::Error: Maximum call stack size exceeded with Nanoc >4.3.7


RubenVerborgh:
Compiling a simple Less stylesheet is broken in Nanoc >4.3.7.

### Steps to reproduce

1. `git clone https://github.com/LinkedDataFragments/Website.git && cd Website `
2. `git checkout cc9bd7f9f6bd1b5b52ab63eed31eb12705a1fbf8`
3. `bundle install`
4. `bundle exec nanoc`

### Expected behavior

It compiles, as it does with Nanoc 4.3.7.

### Actual behavior

```
Less::Error: Maximum call stack size exceeded
```

### Details

Discovered at LinkedDataFragments/Website#1

### Crash log
[crash.log](https://github.com/nanoc/nanoc/files/608519/crash.log.txt)

RubenVerborgh:
Seeing this issue number, I so badly want to find another bug now.

ddfreyne:
`Maximum call stack size exceeded` is an error from the JavaScript runtime which I’ve started encountering in Nanoc as well, but I have never been able to reproduce it locally (only on Travis CI). I’ve tried to figure out where this is coming from, but to no avail.

RubenVerborgh:
I can reproduce on our Ubuntu server; didn't succeed on OSX. Can give you
access if needed.

ddfreyne:
I’ll see about setting up a Vagrant machine here and testing it out that way!

ddfreyne:
Can you do the following for me:

1. Move the contents of `reset.less` into `main.less`
2. Delete `reset.less`
3. Compile

Does the same issue arise in this case as well?

RubenVerborgh:
Tried that, compilation succeeds!

ddfreyne:
Managed to reproduce the problem. This fails:

```
require 'less'

Fiber.new do
  ::Less::Parser.new.parse("p { color: red }").to_css
end.resume
```

This succeeds:

```
require 'less'

::Less::Parser.new.parse("p { color: red }").to_css
```

It appears that the way fibers work is interfering with V8. Further investigation needed…

ddfreyne:
Reported as cowboyd/less.rb#82.

ddfreyne:
Managed to reduce the error to V8:

```
require 'v8'

context = V8::Context.new

Fiber.new do
  context.eval("console.log('hi')")
end.resume
```

Error:

```
[...]

PR #1001, opened 2016-11-26
Run less filter on main fiber

ddfreyne:
This ensures that `Less` is always invoked on the main fiber. Potential fix/workaround for #999.

* Added `Filter#on_main_fiber`, which executes the given block on the main fiber. This is the main mechanism that makes the Less filter work again.

* Converted all Less tests to specs. These are high-level tests (compiling using the CLI) because the Less filter is somewhat annoying to test otherwise (it now requires interaction with the compiler).

* Two new Less specs that did not exist before are skipped for now, as I can’t get them to work. Will investigate in another PR.


ddfreyne:
`less_spec.rb` now passes, but `test_less.rb` fails, with two distinct reasons:

* `Less::Error: Maximum call stack size exceeded`
* `FiberError: dead fiber called`

ddfreyne:
False hope: `less_spec.rb` fails too, sometimes.

ddfreyne:
Not quite working. I reported an issue to `therubyracer`: cowboyd/therubyracer#421.

ddfreyne:
Added a mechanism to make code execute on the main fiber, and made the less filter use it. This seems to work!

ddfreyne:
CC @RubenVerborgh — can you verify that this PR fixes your problem (#999)?

ddfreyne:
Will merge because I cannot reproduce the problem on Travis CI nor locally on a Debian VM.

RubenVerborgh:
That was fast 😉  I'll test later today.

ddfreyne:
The two new skipped Less specs are implemented in #1003.

RubenVerborgh:
Looks good to me.

Issue #1004, opened 2016-11-30
Temporary data is not pruned


ddfreyne:
Nanoc does not remove unused entries from data files in the `tmp/` directory.

These files tend to grow over time and will consume more and more memory, as they’re always fully loaded into memory.

ddfreyne:
A solution that seems obvious to me at the moment is to *not* load data (checksums, compiled content, and dependencies) for items and layouts that don’t exist anymore.

ddfreyne:
#1013 fixes the issue for the compiled content cache. Checksums still need to be pruned. Dependencies are already taken care of.

connorshea:
👏 thanks @ddfreyne :D

ddfreyne:
Checksums are pruned as of #1027, so this issue is resolved.

PR #1013, opened 2016-12-17
Prune old data from compiled content cache

ddfreyne:
This ensures that no cached compiled content is loaded for items that do not exist. This will prevent the compiled content cache from ever growing.

A partial fix for #1004.


ddfreyne:
CC @connorjshea — This is a follow-up on the slow GitLab site compilation bug. (More work necessary to fix it properly, though.)

PR #1025, opened 2016-12-19
Fixed extension_regex when using full identifiers. Fixes #1022

### LINKED ISSUE #1022: YAML metadata parsing filename issue
If a binary file has a somewhat 'complex' filename, nanoc cannot match the binary file with the metadata.

### Steps to reproduce

Given a binary file with a filename like `ubuntu-16.10-server-amd64.iso.torrent`, I would like to create a yaml file with some metadata. According to the [docs](https://nanoc.ws/doc/guides/using-binary-items-effectively/), a yaml file like `ubuntu-16.10-server-amd64.iso.yaml` should work.

However, the only way this works is if I change the binary filename to: `ubuntu-16-10-server-amd64.iso.torrent` (note that I replaced the dot in `16.10` with a dash) and therefore, the  yaml file to `ubuntu-16-10-server-amd64.iso.yaml`.

I know that parsing filenames can be complex (specially considering multiplatform projects), so if there's no way around this, docs should be updated with this case (I can do this :wink:).

Otherwise, thanks for an amazing project!

gpakosz:


ddfreyne:
Replaced by #1059.

PR #1049, opened 2017-01-04
Use ddplugin instead of own plugin implementation

ddfreyne:
Many years ago, I extracted [ddplugin](https://github.com/ddfreyne/ddplugin) out of Nanoc, with the intent of removing the plugin implementation for Nanoc. The time has come to make this happen.

The individual commits make more sense than the entire PR.

**Blocked** because this adds a dependency (ddplugin), which complicates packaging on Debian (and likely others).


ddfreyne:
Merging this. If another 4.4.x release is needed, I’ll create a branch.

PR #1051, opened 2017-01-04
remove require 'parallel' in external links checks

boutil:
remove require 'parallel' in external links checks

### Detailed description

This change proposes to remove the require statement on parallel as it seems it is not used here anymore, and is dropped from the Gemfile.


ddfreyne:
Thanks! Turns out that I had `parallel` in Nanoc’s Gemfile, which I’ll also remove in a bit.

PR #1055, opened 2017-01-10
On-disk compiled content cache

ddfreyne:
Do not load eagerly load compiled content cache data into memory.

### Detailed description

The compiled content cache uses a database called `Nanoc::Int::DDDB`, which is intended for situations where large amounts of data need to be kept out of memory in order to prevent swapping.

The database consists of a data file (with extension `.db`) and an index file (with extension `.idx`):

* The data file consists of a series of records. Each record has a header and a body. The body contains the data of the record, while the header contains the length of the body.

* The index file is a hash table mapping keys to offsets in the data file. It is a Ruby hash that is serialised using Ruby’s marshaling functionality.

The data file is append-only. New entries will be appended to the end, and deleted entries will not be removed from the data file. A database can be compacted, which will replace the data file with a new data file that only contains entries for keys that are also in the index.

Limitations:

* The database can have at most one reader and at most one writer. The database is not thread-safe.

### DIFF
--- a/lib/nanoc/base/repos.rb
+++ b/lib/nanoc/base/repos.rb
@@ -1,4 +1,5 @@
 require_relative 'repos/store'
+require_relative 'repos/dddb'
 
 require_relative 'repos/checksum_store'
 require_relative 'repos/compiled_content_cache'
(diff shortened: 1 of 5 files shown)


ddfreyne:
The usefulness of this is debatable. Keeping this PR open for the time being.

ddfreyne:
The newly added telemetry does not reveal a significant slowdown when using this, which is good. I have not yet verified the memory impact of this, however.

ddfreyne:
Closing for now.

PR #1069, opened 2017-01-16
Do not store temp data per-env

### LINKED ISSUE #1067: 'environment' dependency failure
Items which depends upon environment-specific configuration (in `nanoc.yaml`) are not reliably recompiled when a different environment is chosen.

### Steps to reproduce

Specify a couple environments in `nanoc.yaml`:

~~~yaml
environments:
  default:
    build: development
  prod:
    build: production
~~~

Make an item depend upon the value of `build`. For instance, in `/about.md`:

~~~markdown
Compilation environment: <%= @config[:build] %>
~~~

### Expected behavior

The item (`/about.md`) should be recompiled each time a different environment is chosen. One would expect this output:

~~~shell
$ nanoc compile
Loading site… done
Compiling site…
      create  [0.15s]  output/about/index.html
Site compiled in 0.18s.

$ nanoc -e prod compile
Loading site… done
Compiling site…
      update  [0.06s]  output/about/index.html
Site compiled in 0.06s.

$ nanoc -e default compile
Loading site… done
Compiling site…
      update  [0.06s]  output/about/index.html
Site compiled in 0.06s.

$ nanoc compile
Loading site… done
Compiling site…
Site compiled in 0.01s.
 ~~~

### Actual behavior

The item is correctly re-compiled once, but never again afterward:

~~~shell
$ nanoc compile
Loading site… done
Compiling site…
      create  [0.15s]  output/about/index.html
Site compiled in 0.18s.

$ nanoc -e prod compile
Loading site… done
Compiling site…
      update  [0.06s]  output/about/index.html
Site compiled in 0.06s.

$ nanoc -e default compile
Loading site… done
Compiling site…
Site compiled in 0.01s.

$ nanoc compile
Loading site… done
Compiling site…
Site compiled in 0.01s.
~~~

### Details

~~~shell
$ ruby --version
ruby 2.4.0p0 (2016-12-24 revision 57164) [x86_64-darwin16]
$ nanoc --version
Nanoc 4.5.3 © 2007-2017 Denis Defreyne.
Running ruby 2.4.0 (2016-12-24) on x86_64-darwin16 with RubyGems 2.6.8.
~~~

ddfreyne:
Fixes #1067.

### DIFF
--- a/lib/nanoc/base/repos/checksum_store.rb
+++ b/lib/nanoc/base/repos/checksum_store.rb
@@ -11,8 +11,8 @@ module Nanoc::Int
     c_obj = C::Or[Nanoc::Int::Item, Nanoc::Int::Layout, Nanoc::Int::Configuration, Nanoc::Int::CodeSnippet]
 
     contract C::KeywordArgs[site: C::Maybe[Nanoc::Int::Site], objects: C::IterOf[c_obj]] => C::Any
-    def initialize(site: nil, objects:)
-      super(Nanoc::Int::Store.tmp_path_for(env_name: (site.config.env_name if site), store_name: 'checksums'), 1)
+    def initialize(site: nil, objects:) # rubocop:disable Lint/UnusedMethodArgument
+      super('tmp/checksums', 1)
 
       @objects = objects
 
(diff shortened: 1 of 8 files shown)


Fjan:
This commit has the side effect of doing a full recompile each time the environment changes! In a normal workflow where you fix a typo on staging and then on production you now have to wait for two full recompiles... This makes working with large nanoc sites unbearably slow, can we please revert?

Fjan:
A better solution to prevent issue https://github.com/nanoc/nanoc/issues/1067 would be to do a full recompile only if the output directories for the environments are the same. In normal use you would use two different output directories so the issue would not occur.

ddfreyne:
@Fjan I’ve created a separate issue for this (#1070). This PR can’t be reverted as-is, as it reintroduces a bug.

Issue #1066, opened 2017-01-15
"capturing" outside of :erb


sunshineco:
On my websites, I import several stand-alone HTML files verbatim into `content` as first-class nanoc items which should be incorporated into the website as proper pages, thus laid out and styled the same as the rest of the pages.

For example, a source code project's `README.html` may be imported into the website as nanoc item `/README.html` which is routed to `/documentation/index.html`. Since these are full-blown HTML files, the enclosing `<html>` and `<body>` elements need to be stripped and the body content wrapped instead inside in a `<div>` to become the item's `compiled_content` which is then laid-out via a normal `layout` invocation in a `compile` rule. This stripping of the unwanted HTML elements should be automated (I shouldn't have to do it by hand when importing the HTML file), thus seems a natural fit for a filter. For instance:

~~~ruby
compile '/README.*' do
    filter :imported_html
    layout '/default.*'
    write '/documentation/index.html'
end
~~~

In some cases, these foreign HTML files also have `<style>` and `<script>` elements in `<head>` which need to be incorporated into `<head>` of the item laid out by nanoc. For example, in `layouts/default.html`:

~~~ruby
<html>
    <head>
        ...
        <%= content_for(@item, :head) %>
~~~

The `imported_html` filter looks something like this:

~~~ruby
class ImportedHtml < Nanoc::Filter
  identifiers :imported_html

  def run(content, params={})
    content_for(:head) do
      # Grab <style> and <script> elements from <head>.
      ...
      styles + scripts
    end

    # Grab <body> content.
    ...
    '<div>' + body + '</div>'
  end
end
~~~

The extracted `<style>` and `<script>` elements need to be stored *somewhere* so that they can later be accessed by the layout; since filtering is happening at compile-time, they can't be stored as item attributes (which are *frozen*), thus storing them as `content_for` seems appropriate.

Unfortunately, this doesn't work. `content_for` expects to be called within the context of the `:erb` filter and fails when called outside (say, from some other filter, such as `imported_html`). Specifically, `content_for` fails since there is no `_erbout`. As a work-around, rather than calling `content_for` directly, the filter instead calls `capture_for` which is defined locally as:

~~~ruby
def capture_for(tag)
  _erbout = ''
  content_for(tag) { _erbout << yield }
end
~~~

Although this works, it is *ugly* since it breaks the abstraction of nanoc's `capturing` helper by having special knowledge of its implementation.

[...]

PR #1073, opened 2017-01-19
Allow setting content_for directly

ddfreyne:
Implements #1066.

In addition to what is currently possible,

```ruby
content_for(:head) { _erbout << 'stuff' }
```

it is now possible to pass a string, rather than a block that changes `_erbout`, to `#content_for`:

```ruby
content_for(:head, 'stuff')
```

Note that

```ruby
content_for(:head) { 'stuff' }
```

will (still) not have any effect.

* [x] Add tests
* [x] Allow passing string as first arg
* [x] Refactor

Documentation will be updated afterwards.

### DIFF
--- a/lib/nanoc/helpers/capturing.rb
+++ b/lib/nanoc/helpers/capturing.rb
@@ -66,7 +66,7 @@ module Nanoc::Helpers
 
           unless rep.compiled?
             Fiber.yield(Nanoc::Int::Errors::UnmetDependency.new(rep))
-            return content_for(*args, &block)
+            return run
           end
         end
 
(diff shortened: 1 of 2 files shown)


ddfreyne:
Refactoring idea: passing in `@config` is ugly. Perhaps having `@_context` always available could be useful (and the rendering helper would then always pass `@_context` through).

sunshineco:
Erm, did you mean `s/directory/directly/` in the title of this pull request?

ddfreyne:
@sunshineco Ergh, brain fart. Fixed!

PR #1085, opened 2017-02-17
Allow calling #write twice

### LINKED ISSUE #1037: Cannot use #write more than once
A rule such as

```ruby
compile '/giraffe.*' do
  write '/giraffe.txt'
  write '/giraffe.md'
end
```

will always result in an error saying that there cannot be more than one `:last` snapshot.

ddfreyne:
Fixes #1037. Allows calling `#write` multiple times.

Steps:

* [x] Allow snapshot actions to have multiple names
* [x] Allow snapshot actions to have multiple paths
* [x] Allow item reps to have multiple paths per snapshot
* [x] Merge consecutive snapshot actions into one
* [x] Assign unique names to snapshot actions with no names


ddfreyne:
Forward path references continue to be a pain… e.g.

```ruby
compile '/foo.*' do
  snapshot :raw
  filter :relativize_paths, type: :html unless rep.path.nil?
  write item.identifier.without_ext + '.html'
  snapshot :last
end
```

(`:raw` and `:last` snapshots added for clarity.)

The issue is here is that `rep.path` is hard to predict. If it were to return a string, then the `:last` snapshot will have a path, while the `:raw` snapshot does not. This is in line with expectations.

Now consider this example, where `write` and `filter` are inverted:

```ruby
compile '/foo.*' do
  snapshot :raw
  write item.identifier.without_ext + '.html'
  filter :relativize_paths, type: :html unless rep.path.nil?
  snapshot :last
end
```

If `rep.path` were to return a string, then the `:last` snapshot will not have a path. Inversely, if `rep.path` returns `nil`, then the `:last` snapshot will have a path. This contradicts itself!

ddfreyne:
I’ve decided to not deal with the inconsistency/divergence issue yet. I believe this could be detected (i.e. if the paths have not yet converged after two attempts at calculating them, raise an error).

PR #1098, opened 2017-03-03
Make Filter#depend_on depend on all reps, not just one

### LINKED ISSUE #1093: Filter depend_on dependencies do not define dependency on all reps
When a filter uses `depend_on` to specify dependency on items, not all the dependent item's reps will be compiled before the filter runs. (Relates to #1091)

### Steps to reproduce

```sh
$ nanoc create-site issue-1093
$ cd issue-1093
$ rm -rf layouts content/stylesheet.css

# Naming seems to be relevant for compilation order:
$ touch content/y.png content/z.png

$ cat <<EOF > Rules
#!/usr/bin/env ruby

class TestFilter < Nanoc::Filter
  identifier :test

  def initialize(assigns)
    super

    deps.each do |dep|
      p dep
    end

    depend_on(deps)
  end

  def run(content, _args = {})
    deps.flat_map(&:reps).each do |rep|
      puts "#{rep.path} compiled: #{rep.unwrap.compiled?}"
    end

    content
  end

  private

  def deps
    assigns[:items].find_all('/**/*.png')
  end
end

compile '/**/*.html' do
  filter :test

  write item.identifier.to_s
end

compile '/**/*.png' do
  write @item.identifier.to_s
end

compile '/**/*.png', rep: :foo do
  write @item.identifier.to_s + '.foo'
end

EOF 
```

### Expected behavior

All reps (`:default`, `:foo`) of the PNGs are created before the `:test` filter runs.

Output:
```
Loading site… done
Compiling site…
<Nanoc::ItemWithRepsView identifier=/y.png>
<Nanoc::ItemWithRepsView identifier=/z.png>
      create  [0.00s]  output/y.png
      create  [0.00s]  output/z.png
      create  [0.00s]  output/y.png.foo
      create  [0.00s]  output/z.png.foo
/y.png compiled: true
/y.png.foo compiled: true
/z.png compiled: true
/z.png.foo compiled: true
      create  [0.00s]  output/index.html

Site compiled in 0.04s.
```

### Actual behavior

Only the `:default` rep of the PNGs is created before the `:test` filter runs.

Output:
```
Loading site… done
Compiling site…
<Nanoc::ItemWithRepsView identifier=/y.png>
<Nanoc::ItemWithRepsView identifier=/z.png>
      create  [0.00s]  output/y.png
      create  [0.00s]  output/z.png
/y.png compiled: true
/y.png.foo compiled: false
/z.png compiled: true
/z.png.foo compiled: false
      create  [0.00s]  output/index.html
      create  [0.00s]  output/y.png.foo
      create  [0.00s]  output/z.png.foo

Site compiled in 0.04s.
```

ddfreyne:
Fixes #1093.

### DIFF
--- a/lib/nanoc/base/services/filter.rb
+++ b/lib/nanoc/base/services/filter.rb
@@ -201,17 +201,17 @@ module Nanoc
     #
     # @return [void]
     def depend_on(items)
-      orig_items = items
+      context = @assigns[:item]._context
+
       items = items.map { |i| i.is_a?(Nanoc::ItemWithRepsView) ? i.unwrap : i }
 
       # Notify
-      dependency_tracker = @assigns[:item]._context.dependency_tracker
+      dependency_tracker = context.dependency_tracker
       items.each { |item| dependency_tracker.bounce(item, compiled_content: true) }
 
       # Raise unmet dependency error if necessary
-      items.each do |item|
-        rep = orig_items.sample._context.reps[item].find { |r| !r.compiled? }
-        Fiber.yield(Nanoc::Int::Errors::UnmetDependency.new(rep)) if rep
+      items.flat_map { |i| context.reps[i] }.reject(&:compiled?).each do |rep|
+        Fiber.yield(Nanoc::Int::Errors::UnmetDependency.new(rep))
       end
     end
   end
--- a/spec/nanoc/base/filter_spec.rb
+++ b/spec/nanoc/base/filter_spec.rb
@@ -96,5 +96,32 @@ describe Nanoc::Filter do
         expect(fiber.resume).not_to be_a(Nanoc::Int::Errors::UnmetDependency)
       end
     end
+
+    context 'multiple reps exist' do
+      let(:other_rep) { Nanoc::Int::ItemRep.new(item, :default) }
+
+      before do
+        reps << other_rep
+        rep.compiled = false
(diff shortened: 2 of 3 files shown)


ddfreyne:
FYI, I think that using `Filter#depend_on` is a smell, and Nanoc should not need it. Once Nanoc’s `#raw_path` ensures the file exists (which you probably want if you’re using `#raw_path`), then `#depend_on` might become obsolete.

agross:
How should I use that from my Sass script? Just accessing the path won't make it appear magically during Sass compilation, would it?

ddfreyne:
@agross #1099 will!

agross:
Alright, that might work for single items. Will this PR also fix the depend_on case because this is what I think I need: I use Sass imports for compass sprite generation, hence I have n items that need to be compiled, but I don't want to reference all their paths. Using depend_on with a glob is easier, albeit you calling it a smell ;-)

agross:
Other than that, making reps appear out of thin air is a great addition!

ddfreyne:
> I have n items that need to be compiled, but I don't want to reference all their paths.

These two will have the same effect:

```
depend_on(items)
```

```
items.map(&:reps).flat_map(&:raw_paths)
```

I might make `depend_on` do exactly that behind the scenes!

agross:
:+1:

PR #1113, opened 2017-03-09
Add basic telemetry implementation

ddfreyne:
This generalises the telemetry that’s currently already in Nanoc, to be more usable for other purposes.


ddfreyne:
Cool side-effect: managed to resolve a FIXME about the `--verbose` output being incorrect when compilation gets suspended during filtering.

PR #1121, opened 2017-03-13
Add telemetry for phases

ddfreyne:
This introduces a superclass for all phases, that can notify started/stopped, yielded to/resumed from children, and abortion due to exception:

This will make it easier to add telemetry:

* start: start recording
* yield: stop  recording
* resume: restart recording
* end: stop recording
* abort: stop recording (recording might already be turned off)

This will only be printed if `--verbose` is given at least twice (e.g. `nanoc co -VV`).

screen shot 2017-03-13 at 18 28 36

### DIFF
--- a/lib/nanoc/base/services/compiler/phases.rb
+++ b/lib/nanoc/base/services/compiler/phases.rb
@@ -1,6 +1,8 @@
 module Nanoc::Int::Compiler::Phases
 end
 
+require_relative 'phases/abstract'
+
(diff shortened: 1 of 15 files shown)


ddfreyne:
This refactoring might also make it easier to declare the stack of phases in a nicer way (I’d love to have, literally, `stack = MarkDone >> Write >> Resume >> Cache >> Recalculate`.)

PR #1126, opened 2017-03-18
Memoized and frozen

ddfreyne:
This changes `Memoization` to support frozen objects, and memoizes `IdentifiableCollection#[]`.


ddfreyne:
The coverage failure is addressed in a770a5467e6a67315d26bb2c8186b6f48fa3e758.

PR #1128, opened 2017-03-19
Speed up AttributesModified outdatedness rule

ddfreyne:
This speeds up `AttributesModified` by not repeating the same check in the case of items with multiple reps.

Example on nanoc.ws:

```
       Nanoc::Int::OutdatednessRules::AttributesModified#apply │   91    273   25.0%
```

Hit rate of 25% which isn’t too high, but it is useful.


ddfreyne:
Patch coverage check fails because of an unreachable code patch which deliberately raises an exception.

PR #1140, opened 2017-03-29
Cache rule memory calculator results

ddfreyne:
This prevents the rule memories from being recalculated when it’s not necessary. Calculating the rule memory is fairly expensive, so this leads to a nice speedup.

The design of the way the rule memory is passed around is not ideal. The `ItemRepRouter` is the service that calculates the final rule memory, and it returns that, so that it can be reused elsewhere. This is weird because `ItemRepRouter` both returns something and modifies item reps. Anyway, something to refactor later.

### DIFF
--- a/lib/nanoc/base/services/compiler.rb
+++ b/lib/nanoc/base/services/compiler.rb
@@ -164,7 +164,7 @@ module Nanoc::Int
       builder = Nanoc::Int::ItemRepBuilder.new(
         site, action_provider, @reps
       )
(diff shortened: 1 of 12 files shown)


ddfreyne:
Further refactoring and speed improvement is possible, but I will leave this to a separate PR.

I intend to remove `#outdatedness_reason_for` and `#basic_outdatedness_reason_for` first, in a separate PR, as these methods are no longer useful.

PR #1144, opened 2017-03-29
Perform writes in separate threads

ddfreyne:
**WORK IN PROGRESS** — currently rather broken.

* [ ] Make it actually work properly
  * [ ] Synchronise around stdout output
  * [ ] Synchronise around `output.diff` generation
  * [ ] Call `#finalize`
* [ ] Extract `ThreadPool` abstraction (or whatever)

Preliminary tests suggest a 10% speed improvement.


ddfreyne:
Closing, as it needs far more work.

PR #1150, opened 2017-04-02
Experiment: Mark Nanoc::Filter#initialize as final

ddfreyne:
**Experimental!**

This approach can be used to mark certain methods as final, meaning that they will cause a warning to be printed when attempting to override them.

```
Warning: Nanoc::Filter#initialize is not intended to be overridden. Doing so anyway
can cause issues.
```


ddfreyne:
I’ve decided that even if this works, it’s not the right approach. Methods that aren’t explicitly intended to be overridden, should not be overridden, and classes that are not explicitly intended to be subclassed, should not be subclassed.

PR #1152, opened 2017-04-12
Remove <meta> tag inserted by Nokogiri.

RubenVerborgh:
This pull request removes the `<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">` inserted by Nokogiri, which makes `nanoc check html` fail.

It's a wontfix for Nokogiri (sparklemotion/nokogiri#1008).

### DIFF
--- a/lib/nanoc/filters/relativize_paths.rb
+++ b/lib/nanoc/filters/relativize_paths.rb
@@ -78,7 +78,8 @@ module Nanoc::Filters
         content = content.sub(%r{(<html[^>]+)xmlns="http://www.w3.org/1999/xhtml"}, '\1')
       end
 
-      nokogiri_process(content, selectors, namespaces, klass, type)
+      result = nokogiri_process(content, selectors, namespaces, klass, type)
+      result.sub %r{\s*<meta http-equiv="Content-Type" content="[^"]+">}, ''
     end
 
     def nokogiri_process(content, selectors, namespaces, klass, type)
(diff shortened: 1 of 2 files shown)


ddfreyne:
`<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">` shouldn’t cause the HTML to be invalid. I can’t reproduce the issue locally. Can you show me how it fails?

RubenVerborgh:
In combination with a meta charset tag (as one would typically do in an
HTML5 document), it gives an error.

You can see this with the default Nanoc template.

ddfreyne:
Nokogiri doesn’t support HTML5, so this is expected behavior.

Ideally, Nanoc would use Nokogumbo, but I’ve had trouble getting that one to work, and I believe it’s incompatible with JRuby as well.

If you need the `meta` tag removed, you could write a filter that removes the `<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">`.

RubenVerborgh:
That's what I do now, but it kind of seems a pity that a default Nanoc
filter doesn't support the latest HTML version, and that this filter
produces invalid HTML with the default init site.

RubenVerborgh:
https://github.com/DeSemWeb/ISWC2017/blob/master/lib/better_relativize_paths.rb

gpakosz:
That's because you're running `relativize_paths` after items are laid out right?

RubenVerborgh:
Yes; I also need it to apply to stylesheet and script links from the
template.

ddfreyne:
Take a look at #1153, which uses Nokogumbo (a HTML5 parser) instead of Nokogiri, and thus avoids the `meta` problem.

RubenVerborgh:
Perfect, let's follow up there.

PR #1153, opened 2017-04-14
Support HTML5 in relativize_paths + colorize_syntax

ddfreyne:
Fixes #1152.

* [x] HTML5 support for `:relativize_paths`
* [x] HTML5 support for `:colorize_syntax`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -48,6 +48,7 @@ group :plugins do
   gem 'mime-types'
   gem 'mustache', '~> 1.0'
   gem 'nokogiri', '~> 1.6'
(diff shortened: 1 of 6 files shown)


ddfreyne:
This is a dead end… Nokogiri’s serializer is the one that adds `<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">`, even when using a HTML5 parser.

ddfreyne:
Curious—It seems that `#to_s` does _not_ add the `meta` tag, while `#to_html` does.

ddfreyne:
This is released as part of Nanoc 4.7.6 now.

PR #1158, opened 2017-04-16
Make ActionSequence immutable

ddfreyne:
The goal of this PR is to make `ActionSequence` an immutable object, and move the code that builds action sequences into a separate object (`ActionSequenceBuilder`?), so that computed properties of `ActionSequence` can be cached (`#serialize`, specifically).


ddfreyne:
CC @Fjan — This should bring down compilation speed a little more.

PR #1161, opened 2017-04-19
Configurable meta extension

vsuhachev:
Possibility to redefine metadata file extension for "filesystem" datastore

### Detailed description

added datastore configuratuon param `meta_ext`

### To do

* [ ] Tests
* [ ] Documentation
* [ ] Feature flags

### Related issues

None

### DIFF
--- a/lib/nanoc/data_sources/filesystem.rb
+++ b/lib/nanoc/data_sources/filesystem.rb
@@ -14,9 +14,16 @@ module Nanoc::DataSources
   #
   # The metadata for items and layouts can be stored in a separate file with
   # the same base name but with the `.yaml` extension. If such a file is
-  # found, metadata is read from that file. Alternatively, the content file
-  # itself can start with a metadata section: it can be stored at the top of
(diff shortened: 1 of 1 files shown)


ddfreyne:
Could you elaborate on the use case for making `meta_ext` customisable?

I’d like to limit the number of configuration options in Nanoc to a minimum, and so I’m hesitant to approve this change unless it solves a real problem.

gpakosz:
I second @ddfreyne, beside `.yaml`, or `.yml` what the extension could be?

Maybe nanoc should support these both and that's it.

vsuhachev:
If Filesystem is a private API then this improvement loses his reasons. It looks like I'll have to write my DataSource from scratch

ddfreyne:
@vsuhachev Is there a reason why using `yaml` is not possible for you?

ddfreyne:
FWIW, the [YAML FAQ](http://www.yaml.org/faq.html) indicates `.yaml` as the preferred extension:

> *Is there an official extension for YAML files?*
>
> Please use ".yaml" when possible.

PR #1162, opened 2017-04-19
Original encoding extension point for subclasses

vsuhachev:
Possibility to redefine original encoding for Filesystem class descendants. It may be useful for redefine original encoding for each item

### Detailed description

original_encoding extracted to protected function
fix processing of binary files metadata

### To do

* [ ] Tests
* [ ] Documentation
* [ ] Feature flags

### Related issues

None

### DIFF
--- a/lib/nanoc/data_sources/filesystem.rb
+++ b/lib/nanoc/data_sources/filesystem.rb
@@ -419,6 +419,14 @@ module Nanoc::DataSources
       raise Errors::InvalidMetadata.new(filename, meta.class)
     end
 
+    def original_encoding(filename, data)
+      if @config && @config[:encoding]
+        Encoding.find(@config[:encoding])
+      else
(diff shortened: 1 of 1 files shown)


ddfreyne:
The `Filesystem` class isn’t intended to be subclassed (it’s part of a private API).

Could you elaborate on the use case that you have for this change?

vsuhachev:
The main reason - calculate "original encoding" for each file. I use nanoc to generate html from files with different encodings. I can not change this source data files and want to handle encodings in subclassed Filesystem

ddfreyne:
@vsuhachev Would it be possible for you to do some pre-processing on the files so that they all have the same encoding (UTF-8)?

ddfreyne:
I’ll close this PR for now, but I’m happy to keep the conversation going!

PR #1170, opened 2017-04-22
Improve helper handling in Nanoc::Int::Context

### LINKED ISSUE #1169: `include` approach for activating helpers is problematic
This causes `#path` to be defined everywhere, even when it shouldn’t:

```
module Asdf
  def path
    require 'pry' ; binding.pry
  end
end

include Asdf
```

See deivid-rodriguez/pry-byebug/issues/112.

A workaround is to include the module in Nanoc contexts instead:

```
module Asdf
  def path
    require 'pry' ; binding.pry
  end
end

class Nanoc::Int::Context
  include Asdf
end
```

Nanoc needs to provide an alternative to `include` that does not exhibit this problem.

CC @gpakosz

ddfreyne:
Fixes #1169.

This introduces `use_helper` to load helpers. Helpers loaded with `use_helper` will only be available in contexts where it makes sense, so they don’t pollute the global namespace.

```ruby
module GreetingHelper
  def greeting
    'hello'
  end
end

use_helper GreetingHelper
```

### DIFF
--- a/lib/nanoc/base/entities/code_snippet.rb
+++ b/lib/nanoc/base/entities/code_snippet.rb
@@ -34,6 +34,7 @@ module Nanoc::Int
     #
     # @return [void]
     def load
+      eval('def self.use_helper(mod); Nanoc::Int::Context.instance_eval { include mod }; end', TOPLEVEL_BINDING)
       eval(@data, TOPLEVEL_BINDING, @filename)
       nil
     end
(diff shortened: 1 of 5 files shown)


ddfreyne:
This change makes the helpers not available in the `Rules` file. Making them available there would still be useful.

ddfreyne:
This needs additional tests to verify the behavior of `use_helper`.

PR #1183, opened 2017-05-25
Fix tests on JRuby

ddfreyne:


ddfreyne:
This triggers jruby/jruby#4108.

PR #1184, opened 2017-06-11
Do not use items/layouts (mutable identifier) as hash key

### LINKED ISSUE #1185: Changing identifier causes dep graph inconsistency
Changing the identifier of an item or layout can cause an inconsistency in the dependency graph, as the dependency graph’s vertices use the item/layout’s hash, which will have changed.

### Steps to reproduce

TO DO

### Expected behavior

No crash

### Actual behavior

Crash — a `(document, vertex)` pair exists in the directed graph’s vertex hash, but looking up the vertex for that document yields `nil` instead of the vertex.

### Details

TODO

### Crash log

TODO

ddfreyne:
`Document`’s `@identifier` is mutable, and is therefore not a good candidate for use in a hash key.

* [x] Add test case
* [x] Remove invariant check

Fixes #1185.

### DIFF
--- a/lib/nanoc/base/repos/dependency_store.rb
+++ b/lib/nanoc/base/repos/dependency_store.rb
@@ -14,8 +14,12 @@ module Nanoc::Int
       @items = items
       @layouts = layouts
 
+      @refs2objs = {}
+      items.each   { |o| add_vertex_for(o) }
+      layouts.each { |o| add_vertex_for(o) }
+
       @new_objects = []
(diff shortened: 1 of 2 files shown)


ddfreyne:
Alternative idea: Let the directed graph only contain references, e.g. `[:item, '/foo.md']`.

ddfreyne:
This breaks equality (two objects can be equal even if their hash is different)

ddfreyne:
The directed graph will now store only references, rather than objects themselves.

PR #1191, opened 2017-06-24
Handle on item/layout collections

ddfreyne:
This records dependencies on `@items` and `@layouts`, in order prevent recompiling items that do not (directly or indirectly) depends on the item and layout collections.

Accessing _anything_ from `@items` and `@layouts` will mark an item as outdated when an item or layout is added. Future work includes improving this, by e.g. storing the pattern used by `#[]` and `#find_all`.


ddfreyne:
I missed the main term from the PR title:

> Handle **dependencies** on item/layout collections

not

> Handle on item/layout collections

:|

PR #1196, opened 2017-07-16
Add support for nanoc-rust gem

ddfreyne:
The `nanoc-rust` gem contains parts of Nanoc reimplemented in Rust.

This PR will attempt to load the `nanoc-rust` gem and call `NanocRust.activate!`, which replaces Ruby bits with their Rust equivalent.

⚠️  **Experimental** — although there’s no harm in releasing this, as long as you don’t enable `nanoc-rust` unless you know what you’re doing.

### DIFF
--- a/bin/nanoc
+++ b/bin/nanoc
@@ -2,6 +2,13 @@
 # frozen_string_literal: true
 
 require 'nanoc'
+
+begin
+  require 'nanoc-rust'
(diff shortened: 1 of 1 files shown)


ddfreyne:
Further work will require a mechanism for testing Nanoc against `nanoc-rust`, but this might come later.

PR #1197, opened 2017-07-16
Use Travis CI build stages

ddfreyne:
This is an experiment to divide up the Travis CI build into separate stages:

1. Test
1. Test alternatives (e.g. appraise, with nanoc-rust, …)
1. Test style


ddfreyne:
Not worth it.

PR #1204, opened 2017-08-20
Print site URL when running `nanoc view`

whitequark:
Many terminals (e.g. Konsole) let me open an URL by clicking it, which is quite convenient.

### DIFF
--- a/lib/nanoc/cli/commands/view.rb
+++ b/lib/nanoc/cli/commands/view.rb
@@ -52,6 +52,10 @@ module Nanoc::CLI::Commands
         run Rack::File.new(site.config[:output_dir])
       end.to_app
 
+      # Print a link
+      url = "http://#{options[:host] || 'localhost'}:#{options_for_rack[:Port]}/"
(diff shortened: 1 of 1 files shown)


lib/nanoc/cli/commands/view.rb
```diff
@@ -52,6 +52,10 @@ def run
         run Rack::File.new(site.config[:output_dir])
       end.to_app
 
+      # Print a link
+      url = "http://#{options[:host] || 'localhost'}:#{options_for_rack[:Port]}/"
```
ddfreyne: You could use `options_for_rack[:host]` instead of `options[:host] || 'localhost'`.

This made me wonder whether defaulting to 0.0.0.0, rather than 127.0.0.1, is a sensible default—it probably isn’t. :/

whitequark: Oh huh I had no idea http://0.0.0.0:3000/ would actually work in a browser, turns out it does. Feel free to change to your preferred form.

PR #1205, opened 2017-08-20
Listen on 127.0.0.1, not 0.0.0.0, by default

ddfreyne:
`nanoc view` listens on all network interfaces by default, which is potentially problematic, as exposes the (compiled) Nanoc site to other devices.

This is a potential security issue.


ddfreyne:
CC @whitequark — this is what I mentioned in #1204.

PR #1221, opened 2017-09-21
Add Configuration#dig() and ConfigView#dig()

gpakosz:
Getting closer to winning your eternal gratitude 😄

### DIFF
--- a/lib/nanoc/base/views/config_view.rb
+++ b/lib/nanoc/base/views/config_view.rb
@@ -47,5 +47,11 @@ module Nanoc
       @context.dependency_tracker.bounce(unwrap, attributes: true)
       @config.each(&block)
     end
+
+    # @see Hash#dig
+    def dig(*keys)
+      @context.dependency_tracker.bounce(unwrap, attributes: keys)
+      @config.dig(*keys)
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/nanoc/base/views/config_view.rb
```diff
@@ -47,5 +47,11 @@ def each(&block)
       @context.dependency_tracker.bounce(unwrap, attributes: true)
       @config.each(&block)
     end
+
+    # @see Hash#dig
+    def dig(*keys)
+      @context.dependency_tracker.bounce(unwrap, attributes: keys)
```
ddfreyne: `attributes` is a list of top-level attributes that are touched, so this should be `keys.first` instead. Nanoc only tracks dependencies of keys one level deep.

(`keys` will also work, but generate unnecessary dependencies.)

spec/nanoc/base/views/config_view_spec.rb
```diff
@@ -115,7 +115,25 @@
...
+    end
+
+    context 'with existing keys' do
+      let(:keys) { %i[foo bar] }
+      it { is_expected.to eql(:baz) }
+    end
+
+    context 'with non-existing keys' do
+      let(:keys) { %i[foo baz bar] }
+      it { is_expected.to eql(nil) }
```
ddfreyne: Prefer `be_nil` over `eql(nil)`

ddfreyne:
I’ve applied the changes and merged it into master! Thanks for the quick contribution!

gpakosz:
Thanks for the merge!

PR #1237, opened 2017-10-25
Add :save_with parameter to relativize_paths

gpakosz:
Sometimes, you don't Nokogiri to reformat the document. Typically when you plan to style elements with `white-space: pre;`.

This PR adds a `save_with` parameter to the `relativize_paths` filter to which one can pass e.g. `Nokogiri::XML::Node::SaveOptions::DEFAULT_HTML & ~Nokogiri::XML::Node::SaveOptions::FORMAT` which clears the formatting bit from the default save options.

### DIFF
--- a/lib/nanoc/filters/relativize_paths.rb
+++ b/lib/nanoc/filters/relativize_paths.rb
@@ -61,14 +61,15 @@ module Nanoc::Filters
     end
 
     def relativize_html_like(content, params)
-      selectors  = params.fetch(:select, SELECTORS)
-      namespaces = params.fetch(:namespaces, {})
-      type       = params.fetch(:type)
+      selectors             = params.fetch(:select, SELECTORS)
+      namespaces            = params.fetch(:namespaces, {})
+      type                  = params.fetch(:type)
+      nokogiri_save_options = params.fetch(:nokogiri_save_options, nil)
 
       parser = parser_for(type)
       content = fix_content(content, type)
 
-      nokogiri_process(content, selectors, namespaces, parser, type)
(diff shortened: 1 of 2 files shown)


lib/nanoc/filters/relativize_paths.rb
```diff
@@ -64,11 +64,12 @@ def relativize_html_like(content, params)
       selectors  = params.fetch(:select, SELECTORS)
       namespaces = params.fetch(:namespaces, {})
       type       = params.fetch(:type)
+      save_with  = params.fetch(:save_with, nil)
```
ddfreyne: The name `save_with` is confusing out of context; what about `nokogiri_save_with`?

ddfreyne: This change is missing a test case.

test/filters/test_relativize_paths.rb
```diff
@@ -854,4 +854,34 @@ def test_filter_html_doctype
...
+
+      # Set content
+      raw_content = %(
+<td><span>some</span><span>moderately</span><span>long</span><span>content</span></td>
+)
+
+      # Test
+      nokogiri_save_options = Nokogiri::XML::Node::SaveOptions::DEFAULT_HTML & ~Nokogiri::XML::Node::SaveOptions::FORMAT
+      actual_content = filter.setup_and_run(raw_content.freeze, type: :html, nokogiri_save_options: nokogiri_save_options)
+      assert actual_content == raw_content
```
denis-soundcloud: Prefer `assert_equal(actual_content, raw_content)` — the assertion failure message will be more clear.

gpakosz: Done

PR #1238, opened 2017-10-28
Remove support for Rouge 1.x

ddfreyne:
This removes support for Rouge 1.x, along with the appraisals setup that comes with it.

* Rouge is at version 3.x right now, which means either expanding the setup to include 3.x along with 2.x and 1.x, or dropping support for 1.x, which is unmaintained.

* The appraisals setup is complicated and it’s not clear that `bundle exec rake` does *not* run all tests (it skips Rouge 1.x ones).

@barraq Can you take a look at this and let me know your thoughts/questions/concerns/objections?


spec/nanoc/filters/colorize_syntax/rouge_spec.rb
```diff
@@ -53,12 +58,12 @@ def foo
         let(:output) do
           <<~EOS
             before
-            <pre><code class="language-ruby"><table style="border-spacing: 0"><tbody><tr>
-            <td class="gutter gl" style="text-align: right"><pre class="lineno">1
-            2</pre></td>
-            <td class="code"><pre>  <span class="k">def</span> <span class="nf">foo</span>
-              <span class="k">end</span><span class="w">
-            </span></pre></td>
+            <pre><code class="language-ruby"><table class="rouge-table"><tbody><tr>
```
gpakosz: For the record this is not valid HTML. And is one of the reasons why I departed from `:colorize_syntax` to have my own `:rouge` filter.

ddfreyne:
Will merge this as I see no objection to this.

If the Appraisals setup is needed at a later point in time, it can always be reinstated easily.

PR #1247, opened 2017-11-15
Allow creation of Pattern from Symbol

gpakosz:
A little addition to allow the creation of patterns from symbols, which allows you to write
```
@items[i]
```
where `i` is a `Symbol` instead of having to mess with `#to_s()` at the call site

### DIFF
--- a/lib/nanoc/base/entities/pattern.rb
+++ b/lib/nanoc/base/entities/pattern.rb
@@ -14,6 +14,8 @@ module Nanoc::Int
         Nanoc::Int::StringPattern.new(obj)
       when Regexp
         Nanoc::Int::RegexpPattern.new(obj)
+      when Symbol
+        Nanoc::Int::StringPattern.new(obj.to_s)
       else
         raise ArgumentError, "Do not know how to convert `#{obj.inspect}` into a Nanoc::Pattern"
       end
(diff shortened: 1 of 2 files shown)


ddfreyne:
I really don’t like this idea, because a symbol semantically isn’t the same as a symbol; generating a pattern from a symbol does not sound right to me. I prefer the `#to_s` because it’s more explicit.

gpakosz:
> because a symbol semantically isn't the same as a symbol

🤔

ddfreyne:
Whoops! I meant:

> … because a symbol semantically isn't the same as a *string*

gpakosz:
> … because a symbol semantically isn't the same as a *string*

I'm not sure I understand what you mean. What prompted me to open this PR is that somewhere in a (partial) layout I'm using information that comes from item identifiers or config.

And because of `__nanoc_symbolize_keys_recursively`, data that started their life as strings in YAML metadata end up being symbols down the road.

ddfreyne:
Ah yes, I forgot that a config file like

```yaml
/foo/**/*.md:
  whitelisted: true

/people/**/*.md:
  type: person
```

… would lead to patterns that are symbols.

gpakosz:
OH! You merged fast.

I didn't have the impression I was pushing that hard 😅
Thank you!

Issue #1255, opened 2017-11-28
The erubi filter does not show the location for compilation errors


Fjan:
### Steps to reproduce

1. Create an erb file that parses initially but with an error that will only happen during compilation 
`<%= 'test' # this will parse but breaks at execution time %>`
2. put `filter :erubi` in the Rules file
3. `nanoc compile`
4. the crash output will not show the location of the offending file

### Expected behavior

The crash log should show the item being compiled

### Actual behavior

The crash log shows an incorrect line

### Details

This is trivial to fix, the eval is missing the filename parameter, I will submit a PR shortly.

ddfreyne:
This is fixed in Nanoc 4.8.15 (just released).

PR #1256, opened 2017-11-28
Improve erubi filter error output

Fjan:
this fixes https://github.com/nanoc/nanoc/issues/1255

### DIFF
--- a/lib/nanoc/filters/erubi.rb
+++ b/lib/nanoc/filters/erubi.rb
@@ -23,7 +23,7 @@ module Nanoc::Filters
       assigns_binding = context.get_binding(&proc)
 
       # Get result
-      eval(::Erubi::Engine.new(content, { bufvar: '_erbout', filename: filename }.merge(params)).src, assigns_binding)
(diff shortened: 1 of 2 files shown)


ddfreyne:
Looks good! Can you add a test for this, though? You can use this as a base: https://github.com/nanoc/nanoc/blob/master/test/filters/test_erubi.rb#L24-L37

Fjan:
Ok, I added a test for the specific case I mentioned in the bug report but the test does not create a file on disk so the improvement that the improved error message will show a file name is not really clearly visible

ddfreyne:
Please create a new test, rather than extending the existing one.

For the sake of testing this: you can check whether the exception stack trace contains the item name. That should do it!

Fjan:
I had no luck making an item name appear in the test. Sorry, I really don't have more time to work on this any time soon. As an aside, none of the other filters check for the presence of the item name being present in the error message, so testing this seems more trouble than it's worth for such a small embellishment.

ddfreyne:
All bug fixes need an accompanying test. I’ll work on a fix for this.

ddfreyne:
Replaced by #1259.

PR #1279, opened 2017-12-21
Let #breadcrumbs_trail always find closest parent

### LINKED ISSUE #1278: breadcrumbs_trail misbehaves if item identifiers have more than one "."
Items with more than one "." in their identifier (e.g., "1.6.md") can cause `breadcrumbs_trail` to reference the wrong item.

### Steps to reproduce

Make items like this:

```
/trp/1.5.md
/trp/1.5/foo.md
/trp/1.6.md
/trp/1.6/foo.md
```

The breadcrumbs for one or other of the `foo.md` items will refer to the wrong parent.

### Expected behavior

I'd expect the last element of the breadcrumb trail for an item to always be that item's parent.

### Actual behavior

Sometimes the last element of the breadcrumb trail for an item can be an item with a similar name.

### Details

It looks like the `breadcrumbs_trail` method for full identifiers is stripping extensions for identifier components other than the last: [here](https://github.com/nanoc/nanoc/blob/b322a41d97e6f718d81a8f96aa1a1381620cc92c/nanoc/lib/nanoc/helpers/breadcrumbs.rb#L21)

```ruby
@items[Nanoc::Identifier.new(pr).without_ext + '.*']
```

For example, for `/trp/1.6/software.md`:

* `components` becomes `["", "/trp", "/trp/1.6", "/trp/1.6/software.md"]`

* The identifiers used to construct the prefixes array are: `["/index.*", "/trp.*", "/trp/1.*", "/trp/1.6/software.*"]`

Note that `"/trp/1.*"` can match either `/trp/1.5.md` or `/trp/1.6.md`.

It seems to me as if stripping the extension isn't required here, as long as you _don't_ add the `.*` on to the end of the _last_ identifier in the list: that one will already be present as an explicit extension if we're working with full identifiers.

I could be missing something, though, I am really not a native Ruby programmer (yet).

ddfreyne:
Fixes #1278.

`#breadcrumbs_trail` already supported cases like these:

```
/index.md
/foo.md.erb
/foo/stuff.html

# trail = [/index.md, /foo.md.erb, /foo/stuff.html]
```

but now it will also support

```
/index.md
/1.2.erb
/1.2/stuff.html

# trail = [/index.md, /1.2.erb, /1.2/stuff.html]
```

### DIFF
--- a/nanoc/lib/nanoc/helpers/breadcrumbs.rb
+++ b/nanoc/lib/nanoc/helpers/breadcrumbs.rb
@@ -3,6 +3,34 @@
 module Nanoc::Helpers
   # @see http://nanoc.ws/doc/reference/helpers/#breadcrumbs
   module Breadcrumbs
+    # @api private
+    module Int
+      # e.g. unfold(10.class, &:superclass)
+      # => [Integer, Numeric, Object, BasicObject]
+      def self.unfold(obj, &blk)
+        acc = [obj]
+
+        res = yield(obj)
+        if res
+          acc + unfold(res, &blk)
+        else
+          acc
+        end
+      end
+
+      # e.g. patterns_for_prefix('/foo/1.0')
+      # => ['/foo/1.0.*', '/foo/1.*']
+      def self.patterns_for_prefix(prefix)
+        prefixes =
+          unfold(prefix) do |old_prefix|
+            new_prefix = Nanoc::Identifier.new(old_prefix).without_ext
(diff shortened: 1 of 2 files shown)


iay:
The fancy functional programming stuff is beyond my 2006-vintage understanding of Ruby but the intent seems right. I will try and do a quick test of this tomorrow and let you know whether it works for my case.

iay:
Dropping the modified `breadcrumbs.rb` into my `lib/` directory seems to fix my particular use case.

PR #1296, opened 2018-01-21
Read checks from config too

ddfreyne:
This allows defining deploy checks in the configuration file, rather than the Checks file:

```yaml
checking:
  deploy_checks:
    - internal_links
    - spelling
```

See nanoc/features#27.


ddfreyne:
Idea: Replace “deploy checks” with “default checks”, and let `nanoc check` run them without having to specify `--deploy`.

Issue #1300, opened 2018-01-23
Improve flexibility of Atom feed generator in blogging helper


iay:
The current Atom feed generator uses a single item from the configuration file (`@config[:base_url]`) for three different, logically separate functions:

* The feed's `xml:base` attribute,
* The feed's `<id>`,
* The feed's `<link rel="alternate"/>`.

Only the first of these uses, which specifies the base from which links within content should be interpreted as relative to, can legitimately be regarded as a per-site configuration:

* The `<id>` must be different for different feeds coming from the same site. It would be more sensible for this to default to something related to the `@item` being compiled; it should also be possible to configure this on a per-feed basis.
* The `<link rel="alternate"/>` should be a link to the thing for which this is a feed. It only makes sense for this to be the site's base URL if the whole site is a blog. If the blog is a subsite (and if there are more than one, then at least one of them must not be equivalent to the whole site) then the link should be to that subsite. This should be configurable; the base URL should probably be a prefix for the configured value, allowing no configured value to have the same effect as today.

iay:
This is one where (because I have two blogs on my site, and they are both subsites) my need for this is fairly sharp, and the code looks simple enough that I might be able to propose a PR if you don't think you would have the time to look into it. Let me know if that's the case.

iay:
Another note on this: the W3C validator flags a link like `<link rel="alternate" href="https://iay.org.uk/"/>` with a compatibility recommendation that it should be a same-document reference ([see this](https://validator.w3.org/feed/docs/warning/SameDocumentReference.html).

I think that means that the existing implementation would be improved if it just used `/` here, but perhaps that's a separate issue. Certainly, any new parameter should _not_ be appended to base URL by this reading.

ddfreyne:
Moving this to the [nanoc/features](https://github.com/nanoc/features/issues) repo. (It’s not well-documented that it exists, sorry!)

PR #1301, opened 2018-01-24
Improve flexibility of Atom feed generator

iay:
Allow overriding `<id>` and `<link rel="alternate">` elements.

### Detailed description

Adds two new parameters, `:alt_link` and `:id`, to the `atom_feed` method. These allow the overriding of the `<id>` and `<link rel="alternate">` elements of the generated Atom feed.

Overriding these can result in greater compatibility and also allow for the possibility of multiple feeds per site.

I have manually tested that if you don't supply these new parameters, you get the same output as before. As mentioned in the comments on #1300, I think there's a separate issue that the default for `:alt_link` should probably be something like `"/"` instead of the base URL, but that seems like a separate issue to this enhancement.

### To do

* [ ] Tests
* [ ] Documentation
* [ ] Feature flags
* [ ] …

### Related issues

Would address #1300.

### DIFF
--- a/nanoc/lib/nanoc/helpers/blogging.rb
+++ b/nanoc/lib/nanoc/helpers/blogging.rb
@@ -29,6 +29,8 @@ module Nanoc::Helpers
 
       attr_accessor :config
 
+      attr_accessor :alt_link
+      attr_accessor :id
       attr_accessor :limit
       attr_accessor :relevant_articles
       attr_accessor :preserve_order
@@ -110,14 +112,14 @@ module Nanoc::Helpers
         xml.instruct!
         xml.feed(xmlns: 'http://www.w3.org/2005/Atom', 'xml:base' => root_url) do
           # Add primary attributes
-          xml.id root_url
+          xml.id(id || root_url)
           xml.title title
 
           # Add date
           xml.updated(updated.__nanoc_to_iso8601_time)
 
           # Add links
-          xml.link(rel: 'alternate', href: root_url)
+          xml.link(rel: 'alternate', href: (alt_link || root_url))
           xml.link(rel: 'self',      href: feed_url)
 
           # Add author information
@@ -175,6 +177,8 @@ module Nanoc::Helpers
     # @option params [Boolean] :preserve_order
     # @option params [Proc] :content_proc
     # @option params [Proc] :excerpt_proc
+    # @option params [String] :alt_link
+    # @option params [String] :id
     # @option params [String] :title
     # @option params [String] :author_name
     # @option params [String] :author_uri
@@ -189,6 +193,8 @@ module Nanoc::Helpers
       builder = AtomFeedBuilder.new(@config, @item)
 
       # Fill builder
+      builder.alt_link          = params[:alt_link]
+      builder.id                = params[:id]
       builder.limit             = params[:limit] || 5
       builder.relevant_articles = params[:articles] || articles || []
       builder.preserve_order    = params.fetch(:preserve_order, false)
--- a/nanoc/test/helpers/test_blogging.rb
+++ b/nanoc/test/helpers/test_blogging.rb
@@ -635,4 +635,104 @@ class Nanoc::Helpers::BloggingTest < Nanoc::TestCase
     # Check
     atom_feed
   end
+
+  def test_atom_feed_without_id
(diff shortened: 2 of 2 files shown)


iay:
1. I might take a look this weekend at adding some tests for this, it's about time I learned TDD in Ruby. What's the best reference for the testing framework you're using?

2. I'm not sure how to interpret the CI failures, are they on my side somehow?

3. Do you want me to rebase this against the current HEAD?

iay:
I think I rebased the original commit; CI still failing.

ddfreyne:
@iay The only remaining error that I can see is

```
 1) Failure:
Nanoc::Helpers::BloggingTest#test_atom_feed_without_base_url [/home/travis/build/nanoc/nanoc/nanoc/test/helpers/test_blogging.rb:215]:
[Nanoc::Int::Errors::GenericTrivial] exception expected, not
Class: <NoMethodError>
Message: <"undefined method `+' for nil:NilClass">
---Backtrace---
/home/travis/build/nanoc/nanoc/nanoc/lib/nanoc/helpers/blogging.rb:196:in `atom_feed'
/home/travis/build/nanoc/nanoc/nanoc/test/helpers/test_blogging.rb:216:in `block in test_atom_feed_without_base_url'
---------------
```

… which I believe is related to the change made

iay:
Thanks; I need to fix that but I'll do it in conjunction with learning RSpec and adding some positive tests for the new code. Should get to it over the next couple of weeks (there's really not hurry on this one).

iay:
Reworked to avoid the exception, rebased on latest.

I'm trying to figure out how I would make tests for this change. Am I right in saying that you have both RSpec based tests under `nanoc/spec` and also more conventional unit tests under `nanoc/test`?

It looks like there are no tests for the actual XML build in the RSpec section of the repository, so I'm guessing that it might be easiest to add some tests to the end of `nanoc/test/helpers/test_blogging.rb` and just match expected content within the XML for the item as a string, as for example is being done in for example `test_atom_feed_with_xml_base`. Does that sound right?

iay:
A related question would be whether there was an XML parser available in that test context; it seems like a test that re-parsed the resulting string and then dug into the structure would be less brittle than one involving just regexes.

ddfreyne:
> Am I right in saying that you have both RSpec based tests under `nanoc/spec` and also more conventional unit tests under `nanoc/test`?

Correct. The `test/` directory is older; eventually I hope to have everything migrated to `spec/`, but there’s no rush.

> I'm guessing that it might be easiest to add some tests to the end of `nanoc/test/helpers/test_blogging.rb` and just match expected content within the XML for the item as a string, as for example is being done in `test_atom_feed_with_xml_base`. Does that sound right?

Yes!

ddfreyne:
> A related question would be whether there was an XML parser available in that test context; it seems like a test that re-parsed the resulting string and then dug into the structure would be less brittle than one involving just regexes.

Definitely. Nokogiri likely will be available, so you could do something similar to what’s being done in `test_xml_sitemap.rb`:

```ruby
      doc = Nokogiri::XML(res)
      urlsets = doc.css('> urlset')
      assert_equal 1, urlsets.size
      urls = urlsets.css('> url')
      assert_equal 4, urls.size
      assert_equal 'http://example.com/item-one/a/',   urls[0].css('> loc').inner_text
      assert_equal 'http://example.com/item-one/b/',   urls[1].css('> loc').inner_text
      assert_equal 'http://example.com/item-three/a/', urls[2].css('> loc').inner_text
      assert_equal 'http://example.com/item-three/b/', urls[3].css('> loc').inner_text
      # …
```

iay:
Added unit tests.

So now I know how to use Nokogiri, I guess. Is there anything else you need here?

Also, is there a way to manually run just a single file of tests in isolation? Took a while to get started on this because it takes a while to run everything just to find out that I've written my XPath expressions incorrectly 😄

ddfreyne:
> Also, is there a way to manually run just a single file of tests in isolation?

Yeah, but I just realised that this approach isn’t documented.

From within the `nanoc` subdirectory, either

* (for something in `test`) run `bundle exec m test/path/to/test_something.rb`
* (for something in `spec`) run `bundle exec rspec spec/path/to/something_spec.rb`

PR #1316, opened 2018-02-17
Add mutex to logger

### LINKED ISSUE #1312: Newlines in log can get interleaved on multi-threaded write
Nanoc uses `puts` to write to its log files. That method is currently not atomic in Ruby (it's two separate calls, one to print and one to output a new line). Therefore the recently introduced multi-threaded writing can cause the log output to get interleaved if two threads write at the same time.

### Steps to reproduce

1. Compile a large nanoc site
2. The output will have a few lines with no new line, and some double new lines

### Details

This is actually a bug in Ruby [here](https://bugs.ruby-lang.org/issues/9420) and there are plans to fix it [here](https://bugs.ruby-lang.org/issues/14042). However, there is a very easy workaround, just replace the `puts` call with a `print`. I've verified that this will fix the issue:

`io.print("#{message}\n")   #  in nanoc/cli/logger.rb:70`

ddfreyne:
Fixes #1312.

God knows how to test this.

### DIFF
--- a/nanoc/lib/nanoc/cli/logger.rb
+++ b/nanoc/lib/nanoc/cli/logger.rb
@@ -28,6 +28,7 @@ module Nanoc::CLI
 
     def initialize
       @level = :high
+      @mutex = Mutex.new
     end
 
     # Logs a file-related action.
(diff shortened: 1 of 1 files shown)


Fjan:
Thanks for the fix. However, adding a mutex is a lot heavier than simply using a print statement,  which is atomic by itself.

ddfreyne:
@Fjan I opted for a mutex for two reasons:

1. I’d prefer not to rely on `#print` being atomic — it is not documented to be atomic, and not all Ruby implementations might implement `#print` atomically.

2. Mutexes are so cheap that the slowdown of using mutex is unlikely to be noticeable.

PR #1320, opened 2018-02-21
Handle gcse:search

### LINKED ISSUE #1319: relativize_paths silently, incorrectly transforms `gcse:search`
relativize_paths silently transforms `gcse:search`

### Steps to reproduce

1. Create an html file with Google's GCSE search snippet
    This includes an element for the Google JS to manipulate, `<gcse:search></gcse:search>`
2. Run `nanoc` with rules to apply `relativize_paths` on that html file, with `type: html`
3. Observe output will read `<search></search>`

### Expected behavior

I think that `nanoc` should not be transforming that element at all. However, maybe there's some subtlety I'm missing about strict HTML specification.

Either of the below would be also acceptable:
- `nanoc` should print a warning that the element is transformed -- it can surprise people, after all
- `nanoc` should crashfail the build on the element

### Actual behavior

Instead, `nanoc` succeeds, but changes the element.

### Details

I'm surprised to be the first person reporting this -- maybe not that many people are using GCSE, or using `relativize_paths`, or maybe I'm doing something else wrong.

ddfreyne:
Fixes #1319.

This isn’t pretty, but Nokogiri/libxml don’t offer a clean solution.

### DIFF
--- a/nanoc/lib/nanoc/filters/relativize_paths.rb
+++ b/nanoc/lib/nanoc/filters/relativize_paths.rb
@@ -10,6 +10,8 @@ module Nanoc::Filters
 
     SELECTORS = ['*/@href', '*/@src', 'object/@data', 'param[@name="movie"]/@content', 'form/@action', 'comment()'].freeze
 
+    GCSE_SEARCH_WORKAROUND = 'nanoc__gcse_search__f7ac3462f628a053f86fe6563c0ec98f1fe45cee'
+
     # Relativizes all paths in the given content, which can be HTML, XHTML, XML
     # or CSS. This filter is quite useful if a site needs to be hosted in a
     # subdirectory instead of a subdomain. In HTML, all `href` and `src`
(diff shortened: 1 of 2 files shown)


sirosen:
Thanks for the super-quick turnaround from bug report to fix!
The fix looks reasonable to me (though I agree, not pretty) -- test failure doesn't seem related, maybe an intermittent bug in the testsuite or a ruby2.3 issue.

This is almost the same as the fix we worked out in the project where #1319 surfaced, using a custom filter.

If you want to be *super paranoid*, you could use `ENV['NANOC_GCSE_WORKAROUND'] || 'nanoc__gcse_search_workaround'`.
Probably not necessary, to the point where I won't even advocate for it, but if you worry that this string might actually appear in content in the future, it's an easy out.

ddfreyne:
> test failure doesn't seem related

One of Nanoc’s tests became rather flaky lately. Definitely something I’m looking into.

PR #1325, opened 2018-03-03
Test on Windows using AppVeyor

ddfreyne:
This lets Nanoc be tested on Windows using AppVeyor.

* [x] Test on all supported Ruby versions
* [x] Add AppVeyor badge

There are some things that are currently explicitly *not* supported on Windows:

* Live recompilation
* Symlink to output dir

Additionally, some things are explicitly *not* yet tested on Windows:

* View — needs `fork()`
* Rsync
    * Can be installed with Chocolately (later).
* Redirection handling in `external_links` check
    * Somehow causes the tests to abort. Not sure why!


agross:
You could probably test rsync after installing rsync with Chocolatey: https://chocolatey.org/packages/rsync

ddfreyne:
@agross Yeah, that’ll definitely be useful, though at the moment I’m still trying to get the tests to pass on all recent Ruby versions (2.3 to 2.5).

agross:
Yep, understood. It's always interesting how many platform-specifics are buried in a project. Have been there myself. If you need any help w/ debugging on a Windows machine, let me know.

ddfreyne:
@agross My `appveyor.yml` is quite different from yours. I don’t use `choco` or `ridk`, and also don’t use mingw (no `msys`). Is there a practical difference? What are the advantages/drawbacks of `choco install ruby`?

agross:
@ddfreyne 

> My appveyor.yml is quite different from yours.

I saw that too. It's been my first AppVeyor build, so I'm not sure what's the more "correct" approach.

> I don’t use choco or ridk, and also don’t use mingw (no msys).

It seems AppVeyor has some pre-installed rubies which I wasn't aware of. Hence your `PATH`-setting just works without installing something manually.

> Is there a practical difference? What are the advantages/drawbacks of choco install ruby?

With Chocolatey you get the latest Rubies when they are released on chocolatey.org, not only after  AppVeyor updates its base image. You need the ridk (based on msys) to compile native gems. E.g., nokogumbo will not install without ridk.

There are also differences how the SDK part (now ridk) is installed between Ruby versions. Up until 2.3 you need to install and setup a [different SDK type](https://rubyinstaller.org/downloads/#development-kit-old) than ridk. I never have used the old SDK type with Chocolatey, so I'm not sure how that is handled (but I've seen a Choco package for it). Another disadvantage of Chocolatey is that it takes longer to set up the ruby environment than just using rubies shipped by AppVeyor. 

My appveyor.yaml is basically a reflection of my development machine: You install Ruby and ridk and then `bundle install`.

The `chcp 65001` line is a necessity to get around some nasty external encoding issues when you e.g. capture output from a command and match it against a UTF-8 regex.

Gemfile
```diff
@@ -68,5 +68,6 @@ group :plugins do
   gem 'typogruby'
   gem 'uglifier'
   gem 'w3c_validators'
+  gem 'wdm', '>= 0.1.0' if Gem.win_platform?
```
agross: Instead of the `if` I use this in my Gemfiles:
```rb
install_if(-> { Gem.win_platform? }) do
  gem 'wdm'
end
```

IIRC this leads to a consistent lockfile across OSes, whereas the dep is only installed when needed.

ddfreyne: I picked whatever `listen` recommended.

(I think it’s fine in this case, because the Gemfile.lock isn’t bundled with Nanoc anyway.)

ddfreyne:
I stole the `chcp 65001` from your `appveyor.yml` — until I had that line, I couldn’t figure out what the heck was up with the encoding nor how to fix it 😱

Gemfile
agross: 👍

agross:
> I couldn’t figure out what the heck was up with the encoding nor how to fix it 

Welcome to the beautiful world of Windows (writing this on a Mac 😉)

ddfreyne:
🍾

PR #1335, opened 2018-03-23
Enable passing custom variables to the Haml filter

Ecco:
The Haml filter can now take custom local variables

### Detailed description

```ruby
compile "/**/index.html" do
  layout "/default.haml", assigns: {foo: "bar"}
  # Now, in the layout, one can use the "foo" variable whose value will be "bar"
  write item.identifier
end
```

### DIFF
--- a/nanoc/lib/nanoc/filters/haml.rb
+++ b/nanoc/lib/nanoc/filters/haml.rb
@@ -19,10 +19,11 @@ module Nanoc::Filters
 
       # Create context
       context = ::Nanoc::Int::Context.new(assigns)
+      locals = assigns.merge(params[:assigns] || {})
 
       # Get result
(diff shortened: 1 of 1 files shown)


ddfreyne:
Hi!

I like the idea of Haml locals — it’s indeed not currently supported by Nanoc, but it could be.

Your implementation will work for filters only, though, and not for layouts. For example:

```ruby
compile "/**/stuff.haml" do
  filter :haml assigns: {foo: "bar"}
  write ext: 'html'
end
```

It’s not currently possible to pass locals to layouts. (Though I imagine that could be useful too.) Depending on what you want to achieve, this might not be needed; I personally like to stick things in the frontmatter of an item or a layout and I then rarely have to use locals.

PR #1340, opened 2018-06-09
Add exclude: param to relativize_paths filter

ddfreyne:
Implements nanoc/features/issues/41.

### Detailed description

This allows specifying a list of excludes in the `relativize_paths` filter, e.g.

```ruby
# excludes /foo, /foo/bar, …
# does not exclude /fooooo, /qux/foo, /qux/foo/asdf, …
filter :relativize_paths, type: :html, exclude: '/foo'

# excludes /foo, /foo/asdf, /bar, /bar/asdf, …
# does not exclude /fooooo, /qux/foo, /qux/foo/asdf, /qux/bar, …
filter :relativize_paths, type: :html, exclude: ['/foo', '/bar']

# excludes /foo/bar, /bar, /archive, /blog/archive/something, …
filter :relativize_paths, type: :html, exclude: /ar/
```

It can be used in different ways:

* **string vs. regexp**: When given a regular expression, the path will not be relativized if the regular expression matches. When given a string, the path will not be relativized when the path starts with one or more components that matches the given string.

* **single object vs. array**: The parameter can be either a single string or regexp, or an array of them.

### To do

* [x] Exclude only when fully matching a path component
* [x] Allow specifying string instead of an array of strings
* [x] Allow specifying an (array of) regex instead of an (array of) strings

After merge, before release:

* Document

### DIFF
--- a/nanoc/lib/nanoc/filters/relativize_paths.rb
+++ b/nanoc/lib/nanoc/filters/relativize_paths.rb
@@ -8,6 +8,8 @@ module Nanoc::Filters
     require 'nanoc/helpers/link_to'
     include Nanoc::Helpers::LinkTo
 
+    DDMemoize.activate(self)
+
     SELECTORS = ['*/@href', '*/@src', 'object/@data', 'param[@name="movie"]/@content', 'form/@action', 'comment()'].freeze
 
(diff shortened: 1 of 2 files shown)


DaveEveritt:
👍

PR #1362, opened 2018-10-03
Add title_proc parameter to blogging helper.

th-h:
title_proc can be used to transform the title of each item in an Atom feed, like content_proc or excerpt_proc do for content or excerpt.

Closes nanoc/features#42

### To do

* [ ] Add / expand tests.
* [ ] Add documentation.

Signed-off-by: Thomas Hochstein <thh@inter.net>

PR #1363, opened 2018-10-04
title_proc for the blogging helper 

ddfreyne:
Same as #1362 but with a test.

### Detailed description

title_proc can be used to transform the title of each item in an Atom feed, like content_proc or excerpt_proc do for content or excerpt.

### To do

* [x] Tests
* [ ] Documentation

### Related issues

Closes nanoc/features#42

CC @th-h


ddfreyne:
This is released as part of Nanoc 4.9.8!

PR #1376, opened 2018-11-25
Sass improvements

gpakosz:
This PR adds further improvements to the `:sass` and `:sass_sourcemap` filters.

### Detailed description

- `:sass` and `:sass_sourcemap` filters now require the `:css_path` option, which produces source maps where the optional `"file"` field has a correct value
- the `:sass` filter now supports the generation of inline, base 64 encoded, source maps

* [x] Tests
* [x] Documentation

### Related issues

https://github.com/nanoc/nanoc.ws/pull/227

### DIFF
--- a/nanoc/lib/nanoc/filters/sass.rb
+++ b/nanoc/lib/nanoc/filters/sass.rb
@@ -27,12 +27,23 @@ module Nanoc::Filters
       )
       css_path = options.delete(:css_path) || filter.object_id.to_s
       sourcemap_path = options.delete(:sourcemap_path)
+      if sourcemap_path == :inline
+        sourcemap_path = Nanoc::Int::TempFilenameFactory.instance.create('sass_sourcemap')
+        inline = true
+      end
 
       engine = ::Sass::Engine.new(content, options)
       css, sourcemap = sourcemap_path ? engine.render_with_sourcemap(sourcemap_path) : engine.render
-      sourcemap = sourcemap&.to_json(css_path: css_path, sourcemap_path: sourcemap_path, type: params[:sources_content] ? :inline : :auto)
-      sourcemap = sourcemap&.split("\n")&.reject { |l| l =~ /^\s*"file":\s*"#{filter.object_id.to_s}"\s*$/ }&.join("\n")
-      [css, sourcemap]
+
+      if inline
+        sourcemap = sourcemap&.to_json(css_uri: css_path)
+        encoded = "data:application/json;base64,#{Base64.urlsafe_encode64(sourcemap)}"
+        [css.gsub(%r{^/\*#\s+sourceMappingURL=\s*#{sourcemap_path}\s*\*/$}, "/*# sourceMappingURL=#{encoded} */")]
+      else
+        sourcemap = sourcemap&.to_json(css_path: css_path, sourcemap_path: sourcemap_path, type: params[:sources_content] ? :inline : :auto)
+        sourcemap = sourcemap&.split("\n")&.reject { |l| l =~ /^\s*"file":\s*"#{filter.object_id.to_s}"\s*$/ }&.join("\n")
+        [css, sourcemap]
+      end
     end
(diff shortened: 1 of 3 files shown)


ddfreyne:
Requiring the `:css_path` option will be a backwards-incompatible change. Is there a way to have a default?

nanoc/lib/nanoc/filters/sass.rb
```diff
@@ -25,11 +25,24 @@ def render(filter, rep, content, params = {})
...
+      if sourcemap_path == :inline
+        sourcemap_path = Nanoc::Int::TempFilenameFactory.instance.create('sass_sourcemap')
+        inline = true
+      end
 
       engine = ::Sass::Engine.new(content, options)
       css, sourcemap = sourcemap_path ? engine.render_with_sourcemap(sourcemap_path) : engine.render
-      [css, sourcemap&.to_json(css_uri: rep.path, type: rep.path.nil? ? :inline : :auto)]
+
+      if inline
```
ddfreyne:
```suggestion
      if sourcemap_path == :inline
```

nanoc/lib/nanoc/filters/sass.rb
```diff
@@ -25,11 +25,24 @@ def render(filter, rep, content, params = {})
         filename: rep.item.identifier.to_s,
         cache: false,
       )
+      css_path = options.delete(:css_path)
       sourcemap_path = options.delete(:sourcemap_path)
+      if sourcemap_path == :inline
+        sourcemap_path = Nanoc::Int::TempFilenameFactory.instance.create('sass_sourcemap')
+        inline = true
```
ddfreyne:
```suggestion
```

nanoc/lib/nanoc/filters/sass.rb
```diff
@@ -25,11 +25,24 @@ def render(filter, rep, content, params = {})
...
+      end
 
       engine = ::Sass::Engine.new(content, options)
       css, sourcemap = sourcemap_path ? engine.render_with_sourcemap(sourcemap_path) : engine.render
-      [css, sourcemap&.to_json(css_uri: rep.path, type: rep.path.nil? ? :inline : :auto)]
+
+      if inline
+        require 'base64'
```
ddfreyne: I’d move `require 'base64'` all the way to the top of the file. It’s part of Ruby’s stdlib so it is safe to require there.

nanoc/lib/nanoc/filters/sass.rb
ddfreyne: ignore — this won’t work

nanoc/lib/nanoc/filters/sass.rb
ddfreyne: ignore — this won’t work

gpakosz:
Well I'll think about it, maybe I can do the following: if `css_path` is not given, use a dummy one then remove it from the generated sourcemap instead of letting incorrect information go through

gpakosz:
In fact it has nothing to do with inline source maps. It's just Sass won't let you omit the optional `"file"` entry in the output source map. So the conservative strategy would be to give Sass a dummy value then change that after the source map has been produced

gpakosz:
There, now `css_path` is optional

nanoc/lib/nanoc/filters/sass.rb
ddfreyne: Or even better: stick it somewhere in this list: https://github.com/nanoc/nanoc/blob/master/nanoc/lib/nanoc.rb#L40-L62 (sorted alphabetically)

All of Nanoc’s generic requirements are in there.

It needs to be `require 'base64'` (`requires` is a Nanoc thing — it eventually results in a `require`, but it doesn’t execute `require` until the filter is actually used.)

gpakosz: Done

PR #1379, opened 2018-12-01
Allow Sass to import files even if they don’t have an item

### LINKED ISSUE #1378: Can't compile some scss files importing another scss from node_modules
Some SCSS files aren't compiled with `:sass` filter anymore after #1365 was merged.

### Steps to reproduce

Run this:

```swift
git clone https://github.com/mshibanami/website
cd website
git reset --hard a9fcc7d
npm install
bundle install
bundle exec nanoc
```

### Expected behavior

All SCSS files are compiled without an error.

### Actual behavior

Got this error when compiling [content/assets/sass/revealjs-theme.scss](https://github.com/mshibanami/website/blob/a9fcc7d/content/assets/sass/revealjs-theme.scss):

```
Captain! We’ve been hit!

  ... 36 lines omitted (see crash.log for details)
  9. from /Users/manabu/workspace/website/vendor/bundle/ruby/2.5.0/gems/sass-3.4.25/lib/sass/stack.rb:115:in `with_frame'
  8. from /Users/manabu/workspace/website/vendor/bundle/ruby/2.5.0/gems/sass-3.4.25/lib/sass/stack.rb:79:in `block in with_base'
  7. from /Users/manabu/workspace/website/vendor/bundle/ruby/2.5.0/gems/sass-3.4.25/lib/sass/tree/visitors/perform.rb:162:in `block in visit'
  6. from /Users/manabu/workspace/website/vendor/bundle/ruby/2.5.0/gems/sass-3.4.25/lib/sass/tree/visitors/base.rb:36:in `visit'
  5. from /Users/manabu/workspace/website/vendor/bundle/ruby/2.5.0/gems/sass-3.4.25/lib/sass/tree/visitors/perform.rb:314:in `visit_import'
  4. from /Users/manabu/workspace/website/vendor/bundle/ruby/2.5.0/gems/sass-3.4.25/lib/sass/tree/import_node.rb:37:in `css_import?'
  3. from /Users/manabu/workspace/website/vendor/bundle/ruby/2.5.0/gems/sass-3.4.25/lib/sass/tree/import_node.rb:28:in `imported_file'
  2. from /Users/manabu/workspace/website/vendor/bundle/ruby/2.5.0/gems/sass-3.4.25/lib/sass/tree/import_node.rb:48:in `import'
  1. from /Users/manabu/workspace/website/vendor/bundle/ruby/2.5.0/gems/compass-import-once-1.0.5/lib/compass/import-once/importer.rb:7:in `find_relative'
  /Users/manabu/workspace/website/vendor/bundle/ruby/2.5.0/gems/nanoc-4.10.3/lib/nanoc/filters/sass/importer.rb:24:in `find_relative'

RuntimeError: unable to map /Users/manabu/workspace/website/node_modules/reveal.js/css/theme/template/mixins.scss to any item

Current item: /assets/sass/revealjs-theme.scss (:en representation)

A detailed crash log has been written to ./crash.log.
```

### Details

This problem has been happening after updating the nanoc to `4.9.10`.
I know it was caused by importing SCSS files in `node_modules/reveal.js/css/theme/template/` with `@import`, but I couldn't find the solution due to the lack of my knowledge about the nanoc.

### Crash log

https://travis-ci.org/mshibanami/website/jobs/461728327#L734

ddfreyne:
Fix backwards-incompatible change in 4.10’s Sass filter.

### Detailed description

When no load paths are given to the Sass filter, it should still be able to resolve relative paths (e.g. `../external`) and successfully import. This isn’t currently possible, because the Sass filter assumes that the files have a corresponding Nanoc item.

This change allows the Sass filter to (again) import such files.

### To do

* [x] Tests

### Related issues

Fixes #1378.

### DIFF
--- a/nanoc/lib/nanoc/filters/sass/importer.rb
+++ b/nanoc/lib/nanoc/filters/sass/importer.rb
@@ -20,15 +20,16 @@ module Nanoc::Filters::SassCommon
       return unless raw_filename
 
       item = raw_filename_to_item(raw_filename)
-      # it doesn't make sense to import a file, from Nanoc's content if the corresponding item has been deleted
-      raise "unable to map #{raw_filename} to any item" if item.nil?
 
-      filter.depend_on([item])
+      content = item ? item.raw_content : File.read(raw_filename)
+      filename = item ? item.identifier.to_s : raw_filename
+
+      filter.depend_on([item]) if item
 
       options[:syntax] = syntax
-      options[:filename] = item.identifier.to_s
+      options[:filename] = filename
       options[:importer] = self
-      ::Sass::Engine.new(item.raw_content, options)
+      ::Sass::Engine.new(content, options)
     end
 
     def find(identifier, options)
--- a/nanoc/spec/nanoc/regressions/gh_1378_spec.rb
+++ b/nanoc/spec/nanoc/regressions/gh_1378_spec.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+describe 'GH-1378', site: true, stdio: true do
+  before do
+    FileUtils.mkdir_p('content')
+    File.write('outside.scss', 'p { color: red; }')
+    File.write('content/style.scss', '@import "../outside.scss";')
+
+    File.write('Rules', <<~EOS)
+      compile '/*' do
+        filter :sass, syntax: :scss
+        write ext: 'css'
+      end
+    EOS
+  end
+
+  example do
+    expect { Nanoc::CLI.run([]) }
+      .to change { File.file?('output/style.css') }
+      .from(false)
+      .to(true)
+
+    expect(File.read('output/style.css')).to match(/p\s*{\s*color:\s*red;\s*}/)
+  end
+end
--- a/nanoc/spec/nanoc/filters/sass_spec.rb
+++ b/nanoc/spec/nanoc/filters/sass_spec.rb
@@ -88,11 +88,13 @@ describe Nanoc::Filters::SassCommon do
         rep.raw_paths = rep.paths = { last: [Dir.getwd + '/output/style/main.sass'] }
       end
     end
+
     let(:item_main_sourcemap_rep) do
       Nanoc::Int::ItemRep.new(item_main, :sourcemap).tap do |rep|
         rep.raw_paths = rep.paths = { last: [Dir.getwd + '/output/style/main.sass.map'] }
       end
     end
+
     let(:item_main_view) { Nanoc::CompilationItemView.new(item_main, view_context) }
     let(:item_main_default_rep_view) { Nanoc::CompilationItemRepView.new(item_main_default_rep, view_context) }
     let(:item_main_sourcemap_rep_view) { Nanoc::CompilationItemRepView.new(item_main_sourcemap_rep, view_context) }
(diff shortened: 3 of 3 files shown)


nanoc/lib/nanoc/filters/sass/importer.rb
```diff
@@ -20,15 +20,16 @@ def find_relative(name, base_identifier, options)
       return unless raw_filename
 
       item = raw_filename_to_item(raw_filename)
-      # it doesn't make sense to import a file, from Nanoc's content if the corresponding item has been deleted
-      raise "unable to map #{raw_filename} to any item" if item.nil?
```
gpakosz: Alright, this line indeed makes the compilation stop.

I think you should replace
```
# it doesn't make sense to import a file, from Nanoc's content if the corresponding item has been deleted
raise "unable to map #{raw_filename} to any item" if item.nil?
```
by
```
# the imported file doesn't correspond to a Nanoc item, let next filter in Sass' :load_paths handle it
return if item.nil?
```

And keep the rest of `importer.rb` as it is.

This suggestion alone fixes #1378.

nanoc/lib/nanoc/filters/sass/importer.rb
```diff
@@ -20,15 +20,16 @@ def find_relative(name, base_identifier, options)
...
+      filename = item ? item.identifier.to_s : raw_filename
+
+      filter.depend_on([item]) if item
 
       options[:syntax] = syntax
-      options[:filename] = item.identifier.to_s
+      options[:filename] = filename
       options[:importer] = self
-      ::Sass::Engine.new(item.raw_content, options)
+      ::Sass::Engine.new(content, options)
```
gpakosz: I understand why you're doing it. Indeed it may be desirable to keep backward compatibility and let people `@import` with relative paths where imported files live outside of Nanoc's `content/` directory instead of importing relatively from a configured entry in `load_paths`

nanoc/spec/nanoc/regressions/gh_1378_spec.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+describe 'GH-1378', site: true, stdio: true do
```
gpakosz: Totally cool to learn how to assemble a "real site" instead of having to mock all Nanoc internals 👍

ddfreyne: This is in particular used for regression tests, although I have also started writing integration tests in `spec/nanoc/integration`.

It’s not a substitute for unit tests, though, because the integration/regression tests tend to run slowly and so it’s not great for testing all the different cases.

nanoc/lib/nanoc/filters/sass/importer.rb
ddfreyne: As discussed on Gitter, this won’t work with paths that are relative to the item itself and live outside of the Nanoc content directory.

e.g. current item = `content/style/foo.sass`, path to import = `../../external.sass'

ddfreyne:
As suggested by @gpakosz: a useful refactoring later on might be to have two importer classes — one that deals with Nanoc items exclusively, and one that deals with importing relative files not belonging to a Nanoc item. That’d break up the responsibilities of the two importers a bit more nicely; currently, the single importer does a bit too much.

PR #1385, opened 2018-12-02
Compile in parallel

ddfreyne:
This makes Nanoc compile items in parallel.

### Detailed description

The design:

*   For each outdated item rep, schedule its compilation in a fixed-size thread pool. The thread pool is of a fixed size, so that we can control the effective parallelism.

*   When an outdated item rep fails to be compiled due to a missing dependency, queue its compilation in a separate thread pool without a fixed size (called “extra” for now). This ensures that there cannot be a deadlock due to there not being enough threads. For each element in this “extra” thread pool, there is one item rep in either thread pool that is blocked because of an unmet dependency.

*   The `Resume` phase (now a bad name) notifies the coordinator of failed hard dependencies, while accessing snapshot contents will block until the content is available.

*   One main thread does the coordination through event handling. Eventually, this main thread will also do deadlock detection.

*   To detect deadlocks, Nanoc will record the source and target reps for each `compilation_suspended` notification, and build a graph. If it has cycles, there is a deadlock.

    The current implementation deals with item reps, but that is limited, because there are cases where this approach would detect a false deadlock. For example:

    * item reps R1 and R2
    * R1 has snapshot R1S1 and R1S2
    * R2 has snapshot R2S1 and R2S2
    * R1S2 depends on R2S1
    * R2S2 depends on R1S1

    R1S1 and R2S1 can be compiled without issue. Therefore, both R1S2 and R2S2 can also be compiled. The deadlock detector unfortunately currently detects a deadlock, however.

*   The parallelism in the `Write` phase is removed, as it no longer provides a benefit.

One interesting observation is that the changes are pretty limited:

* The `CompileReps` stage has lots of changes, because that is where the parallelism happens.
* Some stores have basic access mutexes now.
* The `Resume` phase needed only minor changes.
* The `ItemRepSelector` is now obsolete and removed.

### Future work

-   **The `Resume` phase:** It might be possible to remove the `Resume` phase. At the moment, it serves as a way of communicating hard dependencies, since it knows the source rep and the target rep (and target snapshot name), which might not be fully known elsewhere (e.g. the snapshot repo would only know the target rep and target snapshot name).

    The name `Resume` no longer matches the original intent. Perhaps `HardDependencyCatcher` might be a better name. But ideally, we’d get rid of it entirely…

-   **Shared mutable state:** The `dependency_store`, `snapshot_repo`, `compiled_content_cache`, and `outdatedness_store` are all mutable, and shared between threads. Access to them is controlled using mutexes. Rather than having shared mutable state, it might be better to allow the worker threads to communicate with the main thread and request data and modifications to data.

    The snapshot repo (which stores the already-compiled content) uses both mutexes _and_ futures: mutexes to prevent data races, and futures to block until certain content becomes available. The combination of mutexes _and_ futures is a little bizarre.

### To do

* [ ] Tests

* [x] Thread pool

* [ ] Deadlock detection — Thew new implementation can detect dependency cycles, but it will find dependency cycles that are fine (see above for details).

* [ ] Better duration recording — When one rep is blocked on another rep+snapshot, the time spent waiting is currently considered to be part of the rep that requested the content. I suppose it’s not technically _wrong_ for it to be this way, but it might be nice to have the old behavior.

### Related issues

nanoc/features#49


ddfreyne:
Locally, the code seems to be working (apart from deadlock detection), yet on CI is nearly-consistently seems to lock up.

Threads :(

mcorossigo:
Hello @ddfreyne 

> Locally, the code seems to be working

So, is the PR in a usable state for tests/benchmarks?

ddfreyne:
> So, is the PR in a usable state for tests/benchmarks?

Sadly not; the code is quite out of date, tests were never passing properly, and benchmarks showed a 5–10% *decrease* in speed.

ddfreyne:
I will close this for now, and revisit this in the future.

Threads (and related concepts, such as thread pools) are an inherently bad abstraction for parallellism, and I’d like to avoid it entirely, if possible.

Ruby 3’s ractors will likely be helpful here, though I don’t want Nanoc to require Ruby 3 just yet. Hopefully, there is a way to fall back to non-parallel, non-ractor execution when running Ruby 2.x.

PR #1398, opened 2019-01-14
Cache binary content

agross:
First attempt at caching binary items. I'd like to get your feedback on my idea.

Nanoc requiring to recompile binary items is very expensive for us: We use the dependency tracking features of nanoc extensively. The most problematic item is our (Sass) stylesheet that depends on images that are created during compilation. Those images are reps of e.g. SVGs rendered as PNG (via ImageMagick). Whenever we change the Sass stylesheet Nanoc forces us to go through the ImageMagick filter again, even if the change was completely unrelated to images. This takes about 60 seconds.

### Detailed description

My idea was to change the caching infrastructure such that does not break the existing API. I opted for a `CompositeCache` façade that hides the (unchanged) `CompiledContentCache` and adds `BinaryContentCache`.

The `BinaryContentCache` uses a new subdirectory under `/tmp/nanoc/<something>/binary_content` to store cache items. The directory structure is `/tmp/nanoc/<something>/binary_content/<item identifier>/<rep name>/<snapshot name>`. Only binary snapshots will be stored (textual remain in the existing cache). Loading a cache item iterates the `<snapshot name>` files and creates `BinaryContent`s from them.

The `CompiledContentCache` remains unchanged and still stores `TextualContent` and `BinaryContent` as before. Before this patch, if a cache item contained any binary content it was [discarded after retrieval](https://github.com/nanoc/nanoc/blob/master/nanoc/lib/nanoc/base/services/compiler/phases/cache.rb#L36).

The façade delegates to the wrapped caches for the most part. I don't know if there's builtin delegation for a list of wrapped object's in Ruby, so for now it's hard-coded.

For retrieving cache items we take some extra steps:

1. Load cache item `t` from `CompiledContentCache` (returns both textual and binary content)
1. Remove any binary content from `t` (as it would point to the Nanoc `/tmp` directory of the previous run)
1. Load cache item `b` from `BinaryCache` (which uses the file system to generate)
1. Merge `t` and `b` as the cache item
1. Check if the cache item contains all snapshot names as per `rep.snapshot_defs`
1. If yes, the cache item is considered valid
1. If no, the cache item is dropped
1. Compilation proceeds as normal

### To do

* [ ] Verify (or drop) idea
* [ ] Tests (currently failing)
* [ ] Documentation (?)

### Related issues

https://github.com/nanoc/features/issues/20


ddfreyne:
Hey Alex, thanks for the contribution! I might not have time to look into this properly soon, unfortunately, but I like what I see so far.

agross:
Hi there, I got a bit further. The tests are working now but I found a slight issue with the capturing helper. Because capturing seems to inject snapshots like

```rb
{:__capture_blah=>
        #<Nanoc::Core::TextualContent:0x00007fdc3e257638
         @filename=nil,
         @string=#<Nanoc::Core::LazyValue:0x00007fdc3e2572f0 @value={:final=>"Old content"}>}
```

that doesn't make it into the ItemRep's snapshot_defs the compiler thinks that not all reps have been compiled. This stems from the fact that the cache now validates that the cache items' contents actually match the snapshot_defs. And here we have an extra key that essentially invalidates the cache item. My [workaround](https://github.com/nanoc/nanoc/pull/1398/commits/798b6dd9b405f3ee031bcdc16df60d1454cc7293) is to remove such keys before validating, but that feels dirty.

What would you prefer? Drop the validation?

ddfreyne:
@agross I think the validation is useful.

As for naming, I suggest

* rename `CompiledContentCache` to `TextualCompiledContentCache`
* rename `CompositeContentCache` to `CompiledContentCache`
* rename `BinaryContentCache` to `BinaryCompiledContentCache`

… which I think conveys the meaning better.

agross:
Did the rename as suggested.

Specs are now failing because `PostCompileItemRepView` [assumes](https://github.com/nanoc/nanoc/blob/master/nanoc/lib/nanoc/base/views/post_compile_item_rep_view.rb#L13) the cache returns `snapshot_contents` even though the contract of `CompiledContentCache#[]` is `... => C::Maybe[C::HashOf[Symbol => Nanoc::Core::Content]]` with a `Maybe`. I guess until now it never returned `nil` as it would now if the cache item does not validate.

agross:
Fix is in 8ec96a01ed.

agross:
From my point of view everything is speced and tested with one of my projects. The only ugliness is how to handle [this](https://github.com/nanoc/nanoc/pull/1398/commits/5174356b8c99371aaa24ddffd3d2d9c6d3ac640f#diff-c8f30243a3cbb05ecf74432955c88ceeR67).

I'd love to get a get a code review, and perhaps a merge ;-)

ddfreyne:
Hey Alex,

Would you mind if I continue your PR from here? It’s 95% there, but there are some edge cases that I’d like to investigate and some refactoring that I’d like to apply. But the base implementation is great and this enhancement will for sure be welcomed by many people.

(By the way: My internet connection is very spotty the last few days and my ISP can’t say when I will have proper internet again. Communication might be a bit difficult.)

agross:
> Would you mind if I continue your PR from here?

Not at all. Go ahead as you see fit.

> Communication might be a bit difficult

I'm in Berlin next week and the week after - in case you want to meet, talk and/or eat. ;-)

ddfreyne:
> I'm in Berlin next week and the week after - in case you want to meet, talk and/or eat. ;-)

I would love to! Care to send me an email at denis+github@denis.ws so that we can exchange contact details?

ddfreyne:
Closing this in favor of #1408.

PR #1403, opened 2019-01-20
Use Zeitwerk for Core

ddfreyne:
[Zeitwerk](https://github.com/fxn/zeitwerk) is a new Ruby autoloader that seems to be sensible, and avoids the annoyance of having to figure out the order in which `require`s need to happen (because in Ruby, the order matters).

Do not merge until Zeitwerk is 1.0.


fxn:
Awesome diff 😄.

PR #1406, opened 2019-02-09
On EACCES, copy instead of hardlink

### LINKED ISSUE #1405: Nanoc 4.9.2+ crashes when creating hard links on Chrome OS, running inside Termux.
Nanoc 4.9.2+ crashes when creating hard links on Chrome OS, running inside Termux.

### Steps to reproduce

```
rm -rf docs # I've renamed the output directory to 'docs'
nanoc
```

### Expected behavior

Building the nanoc site

### Actual behavior

Crash log:
https://gist.github.com/printfn/6791787b4fb0bb68c86c08b94031a6ea

### Details

The issue seems to be caused by the use of hard links in nanoc 4.9.2. I can reproduce it on 4.9.2, and on the latest 4.11.0. The issue does NOT occur on 4.9.1.

I'm using an Acer Chromebook 15 (CB5-571), running nanoc through Termux from the Play Store. The environment is sandboxed, and the developer mode is disabled.

ddfreyne:
Fixes #1405. (Or rather, works around the issue…)

It is possible for hardlinking to fail when copying works fine. In that case, fall back to copying instead of hardlinking.

To do: Figure out a way to test this.

### Why?

According to [facebook/create-react-app#1832 (comment)](https://github.com/facebook/create-react-app/issues/1832#issuecomment-445500936):

> Due to Android's SELinux policies, you cannot hard link as non-root

### DIFF
--- a/nanoc/lib/nanoc/base/services/item_rep_writer.rb
+++ b/nanoc/lib/nanoc/base/services/item_rep_writer.rb
@@ -61,7 +61,7 @@ module Nanoc
         if is_modified
           begin
             FileUtils.ln(temp_path, raw_path, force: true)
-          rescue Errno::EXDEV
+          rescue Errno::EXDEV, Errno::EACCES
             FileUtils.cp(temp_path, raw_path)
           end
         end


ddfreyne:
CI isn’t my friend today (appears to be testing the wrong commit) but merging anyway (build passes locally) and will verify that `master` passes on CI.

PR #1407, opened 2019-02-11
Infer Sass syntax from item identifier extension, closes #1397

### LINKED ISSUE #1397: Cannot import Sass files by identifier
I use Sass syntax (not Scss). Since the new sourcemap-enabled [importer hard-codes `:scss` syntax](https://github.com/nanoc/nanoc/blob/master/nanoc/lib/nanoc/filters/sass/importer.rb#L45), I cannot use `@import` with item identifiers if the item's content syntax is Sass.

### Steps to reproduce

1. Set up [documented example](https://nanoc.ws/doc/guides/using-common-filters/#using-partials)
1. Make sure the partial contains Sass syntax-specific code
1. Compile

Example repo here: https://github.com/agross/nanoc-sass-partials-by-id

### Expected behavior

Compilation succeeds

### Actual behavior

Compilation fails

```sh
$ nanoc
Loading site… done
Compiling site…
/app.sass:4 DEBUG: Could import by relative path

Captain! We’ve been hit!

Sass::SyntaxError: Invalid CSS after "..."$this-is-sass"": expected "{", was ""

Current item: /app.sass (:default representation)

  0. /_partial.sass:3
  1. /app.sass:6
```

### Details

The build output should contain both `/app.sass:4 DEBUG: Could import by relative path` and `'Could import by item identifier'`.

I need to use `@import "/item/identifier.sass"` instead of plain files because plain files cause a dependency on all items' raw content. This essentially disables Nanoc's caching for the Sass filter output.

### Crash log

https://gist.github.com/agross/ada776af1111de4d54f5f84c81200142

agross:
### Detailed description

Infer Sass or Scss syntax from identifier with fallback to Scss if the identifer has no extension.

This also fixes the specs that always used Scss syntax from `CompiledContentStore`, despite of the created `TextualContent`s with the respective syntax.

### Related issues

#1397, FYI @gpakosz

### DIFF
--- a/nanoc/lib/nanoc/filters/sass/importer.rb
+++ b/nanoc/lib/nanoc/filters/sass/importer.rb
@@ -36,13 +36,9 @@ module Nanoc::Filters::SassCommon
       items = filter.items.find_all(identifier)
       return if items.empty?
 
-      content = if items.size == 1
-                  items.first.compiled_content
-                else
-                  items.map { |item| %(@import "#{item.identifier}";) }.join("\n")
-                end
+      content, syntax = import(items)
 
-      options[:syntax] = :scss
+      options[:syntax] = syntax
       options[:filename] = identifier.to_s
       options[:importer] = self
       ::Sass::Engine.new(content, options)
@@ -84,6 +80,38 @@ module Nanoc::Filters::SassCommon
 
       map = self.class.raw_filename_to_item_map_for_config(filter.config, filter.items)
       map[realpath]
+    end
+
+    private
+
+    def import(items)
+      if items.size == 1
+        import_single(items.first)
+      else
+        import_all(items)
+      end
+    end
+
+    def import_single(item)
+      syntax = if (ext = item.identifier.ext).nil?
+                 nil
+               else
+                 ext.downcase.to_sym
+               end
+
+      [
+        item.compiled_content,
+        syntax,
+      ]
+    end
+
+    def import_all(items)
+      import_all = items.map { |i| %(@import "#{i.identifier}";) }.join("\n")
+
+      [
+        import_all,
+        :scss,
+      ]
     end
   end
 end
--- a/nanoc/spec/nanoc/filters/sass_spec.rb
+++ b/nanoc/spec/nanoc/filters/sass_spec.rb
@@ -68,18 +68,58 @@ describe Nanoc::Filters::SassCommon do
       )
     end
 
-    let(:item_partial) do
+    let(:item_partial_scss) do
       Nanoc::Core::Item.new(
-        content_partial,
+        content_partial_scss,
         { content_filename: 'content/style/_partial.scss' },
         '/style/_partial.scss',
       )
     end
 
-    let(:content_partial) do
+    let(:content_partial_scss) do
       Nanoc::Core::TextualContent.new(
         '* { margin: 0; }',
         filename: File.expand_path('content/style/_partial.scss'),
+      )
+    end
+
+    let(:item_partial_sass) do
+      Nanoc::Core::Item.new(
+        content_partial_sass,
+        { content_filename: 'content/style/_sass-partial.sass' },
+        '/style/_sass-partial.sass',
+      )
+    end
+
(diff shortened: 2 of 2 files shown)


gpakosz:
Hello @agross 👋

I apologize, I got side-tracked and didn't look at #1397. Doing it now

nanoc/lib/nanoc/filters/sass/importer.rb
```diff
@@ -36,13 +36,23 @@ def find(identifier, options)
...
+                            item.compiled_content,
+                          ]
+                        else
+                          import_all = items.map { |i| %(@import "#{i.identifier}";) }.join("\n")
+
+                          [
+                            :scss,
+                            import_all,
+                          ]
+                        end
```
gpakosz: I started changing this file last week but didn't go as far as you did with updating tests.
I wrote something similar:
```
      content, syntax = if items.size == 1
                          [items.first.compiled_content, items.first.ext.downcase.to_sym || :sass]
                        else
                          [items.map { |item| %(@import "#{item.identifier}";) }.join("\n"), :scss]
                        end

      options[:syntax] = syntax
```
Which imho produces a smaller diff than yours.

I chose to make it `content, syntax` because that's somewhat what Sass does, see `find_relative()`.
I chose to make `:sass` the default syntax because that's also what Sass does.

agross: My primary goal is to write readable code, diff size doesn't matter much to me. Git stores snapshots anyways. 😉

The fallback is `:scss` because that's what's the original code was hard-coding. 

I needed to change the specs because they were failing now that Sass compiles the inferred syntax, but the content served by the cache was Scss. 

I'm open to exchanging `syntax` and `content` if that is preferred.

gpakosz: Readable is a subjective matter. When I talk about minimizing the diff, I talk about the cognitive load of inspecting the diff, not Git's way of storing things. I let @ddfreyne decide.

Having `content, syntax` is symetric with what Sass does and what `#find_relative()` does.

As far as the current behavior of assuming `:scss`, it's a bug and I vaguely recall my first implementation before review inspected item's extension to decide the syntax. Maybe we could `#raise()` telling we don't know which syntax to pick?

ddfreyne: I don’t have much of an opinion here — if the tests pass and Rubocop is happy with it, then I’m happy.

nanoc/lib/nanoc/filters/sass/importer.rb
```diff
@@ -36,13 +36,23 @@ def find(identifier, options)
...
-      content = if items.size == 1
-                  items.first.compiled_content
-                else
-                  items.map { |item| %(@import "#{item.identifier}";) }.join("\n")
-                end
+      syntax, content = if items.size == 1
+                          item = items.first
+
+                          [
+                            item.identifier.ext.downcase.to_sym || :scss,
```
ddfreyne: I think the `|| :scss` case is not tested, and likely not correct either (if there is no extension, `.downcase` would raise a NoMethodError I expect).

agross: Good catch! In the extensionless case Nanoc now specifies `syntax = nil` which makes Sass use its internal default (which is Sass).

nanoc/lib/nanoc/filters/sass/importer.rb
agross: @gpakosz Made the changes as requested. I also added some new specs round importing partials.

gpakosz: @agross Thanks for the update. Well my very own opinion is that I wouldn't have traded 5 lines of code in context for 3 additional private functions (see http://number-none.com/blow/john_carmack_on_inlined_code.html). Undersanting what's going on imho now requires chasing tiny functions doing little meaningful work.

So yeah I personally would have written it differently. But if both Rubocop and @ddfreyne are happy then I'm happy too!

PR #1411, opened 2019-02-16
Let Zeitwerk ignore anything in core/core_ext

fxn:
I don't know if the original listing file by file was intentional (in which case please just disregard this PR), but it is possible to let Zeitwerk ignore directories too.


ddfreyne:
Thank you, that does make it a bit nicer!

Thank you for your work on Zeitwerk. I’ve always been wary of autoloading mechanisms, but Zeitwerk is the first one that really gets it right.

PR #1417, opened 2019-04-27
Set a User-Agent for HTTP requests from :elinks

da2x:
### Detailed description

* The default “Ruby” User-Agent is blocked everywhere so replace it.
* Include “Mozilla/5.0” to pass through a lot of blocking filters and because it’s required due to the legacy of the web.
* Include Nanoc software name and version to identify the software making the request.
* Describe the purpose of the request to help webmasters make informed blocking decisions.

Proposed User-Agent string:

    Mozilla/5.0 Nanoc/4.11.2 (link rot checker)

### To do

* [ ] Agree on User-Agent string.
* [ ] Add to change log.

## More details

In my own set of 1800 links, this change reduces the number of HTTP 403 Forbidden responses from 482 to 17. (Most of these are  probably hosted by a single large provider like Cloudflare.) This could probably be reduced even further by faking a mainstream web browser’s User-Agent string. However, I’d rather live with a handful of blocked tests than start faking User-Agents.

### DIFF
--- a/nanoc/lib/nanoc/checking/checks/external_links.rb
+++ b/nanoc/lib/nanoc/checking/checks/external_links.rb
@@ -125,6 +125,7 @@ module ::Nanoc::Checking::Checks
 
     def request_url_once(url)
       req = Net::HTTP::Get.new(path_for_url(url))
+      req['User-Agent'] = "Mozilla/5.0 Nanoc/#{Nanoc::VERSION} (link rot checker)"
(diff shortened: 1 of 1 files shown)


ddfreyne:
Ooh, I like it. I think the User Agent that you propose is good, and I suppose it can be changed in the future if needed.

Could be useful to have a test for this, but probably not important. (I also realised that I haven’t really set up proper testing for the external links checker yet anyway…)

PR #1422, opened 2019-04-27
Collect even more links from even more sources

da2x:
### Detailed description

* Support area, base, blockquote, object, and source elements.
* Support srcset attribute on img and source elements (for responsive images).
* Support poster attribute on the video element.
* Support ping attribute on a and area elements.
* Support HTML+RDFa global about and resource attributes.

### To do

* [x] A test including an image srcset attribute (I still can’t run the tests locally)

### DIFF
--- a/nanoc/test/checking/checks/test_internal_links.rb
+++ b/nanoc/test/checking/checks/test_internal_links.rb
@@ -3,13 +3,12 @@
 require 'helper'
 
 class Nanoc::Checking::Checks::InternalLinksTest < Nanoc::TestCase
-  def test_run
+  def test_no_issues
     with_site do |site|
       # Create files
       FileUtils.mkdir_p('output')
-      FileUtils.mkdir_p('output/stuff')
-      File.open('output/foo.txt',  'w') { |io| io.write('<a href="/broken">broken</a>') }
-      File.open('output/bar.html', 'w') { |io| io.write('<a href="/foo.txt">not broken</a>') }
+      File.open('output/foo.xhtml', 'w') { |io| io.write('<a href="/bar.html">not broken</a>') }
+      File.open('output/bar.html', 'w') { |io| io.write('<a href="/foo.xhtml">not broken</a>') }
 
       # Create check
       check = Nanoc::Checking::Checks::InternalLinks.create(site)
@@ -17,6 +16,21 @@ class Nanoc::Checking::Checks::InternalLinksTest < Nanoc::TestCase
 
       # Test
       assert check.issues.empty?
+    end
+  end
+
+  def test_has_issues
+    with_site do |site|
+      # Create files
+      FileUtils.mkdir_p('output')
+      File.open('output/foo.html', 'w') { |io| io.write('<a href="/broken">broken</a>') }
+
+      # Create check
+      check = Nanoc::Checking::Checks::InternalLinks.create(site)
+      check.run
+
+      # Test
+      refute check.issues.empty?
     end
   end
 
@@ -38,7 +52,6 @@ class Nanoc::Checking::Checks::InternalLinksTest < Nanoc::TestCase
   def test_valid?
     with_site do |site|
       # Create files
-      FileUtils.mkdir_p('output')
       FileUtils.mkdir_p('output/stuff')
       File.open('output/origin',     'w') { |io| io.write('hi') }
       File.open('output/foo',        'w') { |io| io.write('hi') }
@@ -48,12 +61,12 @@ class Nanoc::Checking::Checks::InternalLinksTest < Nanoc::TestCase
       check = Nanoc::Checking::Checks::InternalLinks.create(site)
 
       # Test
-      assert check.send(:valid?, 'foo',         'output/origin')
-      assert check.send(:valid?, 'origin',      'output/origin')
-      assert check.send(:valid?, 'stuff/blah',  'output/origin')
-      assert check.send(:valid?, '/foo',        'output/origin')
-      assert check.send(:valid?, '/origin',     'output/origin')
-      assert check.send(:valid?, '/stuff/blah', 'output/origin')
+      assert check.send(:valid?, path_to_file_uri('foo', site),         'output/origin')
+      assert check.send(:valid?, path_to_file_uri('origin', site),      'output/origin')
+      assert check.send(:valid?, path_to_file_uri('stuff/blah', site),  'output/origin')
+      assert check.send(:valid?, path_to_file_uri('/foo', site),        'output/origin')
+      assert check.send(:valid?, path_to_file_uri('/origin', site),     'output/origin')
+      assert check.send(:valid?, path_to_file_uri('/stuff/blah', site), 'output/origin')
     end
   end
 
@@ -64,7 +77,8 @@ class Nanoc::Checking::Checks::InternalLinksTest < Nanoc::TestCase
 
       check = Nanoc::Checking::Checks::InternalLinks.create(site)
 
-      assert check.send(:valid?, 'stuff/right?foo=123', 'output/origin')
+      assert check.send(:valid?, '/stuff/right?foo=123', 'output/origin')
+      assert check.send(:valid?, 'stuff/right?foo=456', 'output/origin')
       refute check.send(:valid?, 'stuff/wrong?foo=123', 'output/origin')
     end
   end
@@ -75,9 +89,9 @@ class Nanoc::Checking::Checks::InternalLinksTest < Nanoc::TestCase
 
       check = Nanoc::Checking::Checks::InternalLinks.create(site)
 
-      assert check.send(:valid?, '/excluded1', 'output/origin')
-      assert check.send(:valid?, '/excluded2', 'output/origin')
-      assert !check.send(:valid?, '/excluded_not', 'output/origin')
+      assert check.send(:valid?, path_to_file_uri('/excluded1', site), 'output/origin')
+      assert check.send(:valid?, path_to_file_uri('/excluded2', site), 'output/origin')
+      refute check.send(:valid?, path_to_file_uri('/excluded_not', site), 'output/origin')
     end
   end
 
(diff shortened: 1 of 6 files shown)


ddfreyne:
For the tests: I did a bit of digging and found

- https://stackoverflow.com/questions/51661931/rails-s-command-issue/51662528
- https://github.com/cowboyd/therubyracer/issues/453

I think with `therubyracer` commented-out in the Gemfile, the tests might start to work. (Some other tests might fail, but at least it’s a start.)

da2x:
So the mixed content filter depends on `resource_uris_in_file` only checking document embeds. I didn’t quite understand the difference between it and `hrefs_in_file` when looking at the code. OK, so all elements but `a` and `area` (unless they point to a document under the `base_uri` or the `base` element’s URI) can affect mixed content and the `GLOBAL_ATTR`s don’t either.

To get the mixed content tester working the remaining issues are:

 - [ ] Does Nokogiri handle and return absolute URIs for document with a base URI or does Nanoc needto handle that?
 - [ ] Return `a` and `area` elements linking to the base_uri in the site configuration in `resource_uris_in_file`.
 - [ ] Filter out the about and resource attributes when running the mixed content test.

I’m inclined to say this is out of scope for this PR as the mixed content checker didn’t really do it’s job too well prior to this change; and this change will let the internal and external link checkers do a way better job.

ddfreyne:
@da2x I’m rather confused by the original code myself now. :( (I assure you that this does not happen very often.)

`LinkCollector` would also better be named `URLCollector` I believe, since it collects URLs (for things to check). It might be pedantic, but URLs ≠ links, and I think the distinction is important in this case.

ddfreyne:
More potential pedantry: tags ≠ elements, even though the LinkCollector refers to elements as tags.

da2x:
It’s actually a URI collector and not a URL collector to be even more pedantic as it picks up on all URI schemes.

Anyhow, I’m unsure how to proceed with this. How about dropping the mixed content checker altogether? The current version wouldn’t pick up on a bunch of stuff yet I can’t find any bugs relating to it. To me, this suggests that no one is using it.

The mixed content checker could also be changed to check all links to the site conf base_uri domain under the http scheme. This would also pick up on internal links that could downgrade security. This wouldn’t pick up on external embeds, however.

ddfreyne:
I’d like to keep the `mixed_content` check; it’s mentioned on the web site (https://nanoc.ws/doc/testing/ near the bottom).

Would it make sense to split `LinkCollector` into one class for collecting URIs for the link checker, and another one for mixed content? The mixed-content one might still not be working great, but at least you’d have the freedom to structure `URICollector` the way you want and let it do what you need.

da2x:
I’ll change things so `resource_uris_in_file` (for `:mixed_content`) returns embeds and `hrefs_in_file` returns everything (for `:ilinks` and `:elinks`). I’ll then fix al three checkers to use the right one. As a bonus this should speed up each check as each check will only need one pass through all the output files.

da2x:
Current status: Everything should be working including the mixed content checker (except downgrade links but those were already not tested). The internal_link checker now handles deeply-nested relativized links and more variations of same-directory file links.

I still can’t run the tests locally even when commenting out therubyracer. I’m on Fedora so gcc hardening is applied, however. I don’t want to blindly writing more tests, however.

@ddfreyne, what is the failing test? I can’t tell from the output. The same tests failed in #1424.

I believe this should be at a mergeable state now.

da2x:
So there was a number of issues higher than zero. I’ve addressed those now and added tests for new link collection sources, and changed the ilinks tests to work with URIs from the link collector.

I don’t know whats with `test_nanoc_cruby25` and the CCI output doesn’t give me much to work with. `test_nanoc_cruby24` times out after ten minutes but there are no details on what it got hung up on. Both also failed in the https://github.com/nanoc/nanoc/pull/1424 hotfix with the same CCI output.

ddfreyne:
Yeah, the tests on CI started to become flaky and I have not investigated in detail yet. I plan on doing that soon!

nanoc/test/extra/test_link_collector.rb
```diff
@@ -16,95 +19,157 @@ def test_all
...
     refute_includes hrefs, 'http://example.com/'
     refute_includes hrefs, 'https://example.com/'
-    assert_includes hrefs, 'stuff/'
+    refute_includes hrefs, 'https://nanoc.ws'
     refute_includes hrefs, 'mailto:bob@example.com'
-    assert_includes hrefs, '../stuff'
-    assert_includes hrefs, '/stuff'
+  end
+
+  def collect_links_from_space_separated_lists
```
ddfreyne: Test method names have to start with `test_`, or they won’t be run

nanoc/test/extra/test_link_collector.rb
```diff
@@ -16,95 +19,157 @@ def test_all
...
+    assert_includes hrefs, path_to_file_uri('ping1', Dir.pwd)
+    assert_includes hrefs, path_to_file_uri('ping2', Dir.pwd)
+    assert_includes hrefs, path_to_file_uri('ping5', Dir.pwd)
+    assert_includes hrefs, path_to_file_uri('ping3', Dir.pwd)
+    assert_includes hrefs, 'http://example.com/ping4'
+    refute_includes hrefs, nil
+    refute_includes hrefs, path_to_file_uri('/', Dir.pwd)
+  end
+
+  def collects_exotic_links
```
ddfreyne: Same here: add `test_` prefix

nanoc/test/checking/checks/test_internal_links.rb
```diff
@@ -111,8 +125,22 @@ def test_unescape_url
...
       check = Nanoc::Checking::Checks::InternalLinks.create(site)
 
-      assert check.send(:valid?, 'stuff/right%20foo', 'output/origin')
-      refute check.send(:valid?, 'stuff/wrong%20foo', 'output/origin')
+      assert check.send(:valid?, path_to_file_uri('stuff/right%20foo', site), 'output/origin')
+      refute check.send(:valid?, path_to_file_uri('stuff/wrong%20foo', site), 'output/origin')
+    end
+  end
+
+  def deeply_nesten_relatives
```
ddfreyne: Same here: add `test_` prefix

(I eventually want to move fully to RSpec, but it’s a slow work in progress.)

da2x:
OK, one more go at this. test_nanoc_cruby24 and test_nanoc_cruby25 still not giving me any usable information.

ddfreyne:
Whoa, `test_nanoc_cruby24` and `test_nanoc_cruby25` don’t seem to do anything… but not consistently. I’ve rescheduled test runs because it seems that _sometimes_, they do work. I haven’t been able to figure out why, and this is the first time I notice it.

da2x:
No go. Try again?

ddfreyne:
Very odd — it seems that this particular branch fails consistently, but only on CI. I can’t make it fail locally.

I created a branch to test out this weird behavior (#1433), but that branch also does not fail, and I can’t make it fail even when re-running over and over.

da2x:
Lets run this branch with trace then.

TODO: Don’t merge without reverting https://github.com/nanoc/nanoc/pull/1422/commits/05c5d193841bbdb2a9c54c14bbd0f79b7346c40c

da2x:
OK, so that gave us no new information.

da2x:
I rebased this branch again to no avail. I managed to setup a Ruby 2.5 environment locally and the tests ran just fine.

@ddfreyne Any idea why the #1424 branch failed in the same way? There is really no reason why that branch should have fail. Could you try restarting the tests in that branch for comparison?

Could it be that CCI just doesn’t like branches created by me? :wink:

ddfreyne:
I’m also stumped. I’ve never seen this behavior — I cannot figure out why `rake` would silently exit.

ddfreyne:
I SSH’d into the test container (which you can do with CircleCI), changed the test configuration to be verbose, and got this:

```
$ bundle exec rake test --verbose --trace
** Invoke test (first_time)
** Invoke spec (first_time)
** Execute spec
/usr/local/bin/ruby -I/home/circleci/project/vendor/bundle/gems/rspec-core-3.8.0/lib:/home/circleci/project/vendor/bundle/gems/rspec-support-3.8.0/lib /home/circleci/project/vendor/bundle/gems/rspec-core-3.8.0/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb

  1) Nanoc::Checking::Checks::ExternalLinks found has no issues
     Failure/Error: base_uri = URI::File.build([nil, filename])

     NameError:
       uninitialized constant URI::File
     # ./lib/nanoc/extra/link_collector.rb:88:in `uris_in_file'
     # ./lib/nanoc/extra/link_collector.rb:64:in `hrefs_in_file'
     # ./lib/nanoc/extra/link_collector.rb:43:in `block in filenames_per_href'
     # ./lib/nanoc/extra/link_collector.rb:78:in `block in grouped_filenames'
     # ./lib/nanoc/extra/link_collector.rb:77:in `each'
     # ./lib/nanoc/extra/link_collector.rb:77:in `grouped_filenames'
     # ./lib/nanoc/extra/link_collector.rb:43:in `filenames_per_href'
     # ./lib/nanoc/checking/checks/external_links.rb:14:in `run'
     # ./spec/nanoc/checking/checks/external_links_spec.rb:41:in `block (3 levels) in <top (required)>'
     # /home/circleci/project/common/spec/spec_helper_foot.rb:40:in `block (2 levels) in <top (required)>'
     # /home/circleci/project/common/spec/spec_helper_foot_core.rb:31:in `block (4 levels) in <top (required)>'
     # /home/circleci/project/common/spec/spec_helper_foot_core.rb:6:in `__nanoc_core_chdir'
     # /home/circleci/project/common/spec/spec_helper_foot_core.rb:31:in `block (3 levels) in <top (required)>'
     # /home/circleci/project/common/spec/spec_helper_foot_core.rb:30:in `block (2 levels) in <top (required)>'
```

… and a handful of similar failures.

* * *

FYI, I changed the `:spec` task in Rakefile to say

```
RSpec::Core::RakeTask.new(:spec) do |t|
  t.verbose = true
end
```

ddfreyne:
`uninitialized constant URI::File` does not make sense to me, because Nanoc does require `uri`, and then `URI::File` becomes available…

ddfreyne:
@da2x Ohhhh, `URI::File` does not exist in Ruby 2.5.x — it’s new in Ruby 2.6. That would explain…

ddfreyne:
The “test summary” is supposed to list all the failed tests, but that is clearly not quite working the way I expected…

da2x:
Thanks, trying URI::Generic instead. Seems to handle Windows paths as well so should be fine.

What about test_nanoc_cruby24?

ddfreyne:
Can you apply the changes from #1434 to your branch? That should get you actual test results!

da2x:
All green!

…

So there is a small behavioral difference in Ruby2.4 compared to 2.5 and 2.6 that I’ll just leave in. `URI::Generic` isn’t aware of the file URI scheme and blindly merges the protocol-level slashes. So it returns `file:/X` instead of `file:///X`  and `file:/c:/X` instead of `file://c:/X` on Windows. Everything works, notably the improved handling of deeply nested relative paths, but you’ll need to be aware of the difference when handling raw output from the LinkCollector. I’ve changed one test to account for that difference.

Issue #1418, opened 2019-04-27
:external_links checker sends too many requests to the same hosts


da2x:
### Steps to reproduce

1. Create some items that links multiple times to different websites. (item1 links ten times to example.com, item2 links ten times to example.org, etc.)
2. Run` nanoc check :elinks`

### Actual behavior

Nanoc sends requests in the order the links where discovered and ends up sending multiple requests to the same origin in a short amount of time. This can also happens when frequently linking to the same website from multiple items.

Results in HTTP 429 Too Many Requests rate-limiting responses.

### Expected behavior

Nanoc should avoid sending repeated requests to the same origin multiple times in a row. Introducing a delay between requests is undesirable, but shuffling all the collected links from all items in a way that spreads out requests to the same origins would be great. Not sure how to best do that, though.

ddfreyne:
I think shuffling links (with `Array#shuffle`) would be a good first step.

A HTTP 429 Too Many Requests could include a `Retry-After`, which Nanoc could use to figure out how long to wait before retrying requests.

ddfreyne:
Alternatively, it might be interesting to let Nanoc delegate to @gjtorikian’s [html-proofer](https://github.com/gjtorikian/html-proofer). I don’t know how feature-complete its link checker is, but I’m all for reusing rather than rebuilding.

iay:
I've been having some similar problems with my site; my approach so far as been to either link to sites that are less picky, or disabling the check in configuration. I'm honestly not sure whether in the cases I've seen that spacing things out would make much difference... the whole process doesn't take all that long.

Longer term, I think it might be worth refactoring the link checker to allow user-defined strategy to be injected. I don't recall if the current code ignores duplicates or not, but that would be the sort of canonical example of what I have in mind.

da2x:
This strategy should work well. I’ve got no idea how much this would impact performance, though. —and I don’t volunteer to implement it. 😛

1. Sort all links into arrays by origin (see below)
2. Pop one item from the first array and check it
3. Pop one item from the next array and check that
4. Repeat 3 until you reach the end
5. Go to 2

```
{
  'https://example.com': [
    '/document1',
    '/document2'
  ],
  'http://example.org': [
    '/document3',
  ],
  'https://example.net': [
    '/document4',
    '/document5',
    '/document6'
  ]
}
```

This should check in order (the spread improves with more links):

```
https://example.com/document1
http://example.org/document3
https://example.net/document4
https://example.com/document2
[...]

PR #1429, opened 2019-05-01
Shuffle external links list (:elinks checker)

da2x:
### Detailed description

Shuffle the list of links to test. The set will normally be sorted per item they were collected in and in many situation trend towards having the same origin repeated multiple times in a row.

Same-origin distance distribution (number of other links between links to the same origin) in my set of 2000+ links from 500 items **before** to change:

    Min.   1st Qu.  Median    Mean   3rd Qu.    Max. 
     1.0       2.0    10.0   135.8     103.8  2089.0 

Same-origin distance distribution **after** to change:

    Min.   1st Qu.  Median    Mean   3rd Qu.    Max. 
     1.0      15.0    52.5   207.0     202.0  1842.0

It’s not perfect but it’s an improvement.

### Related issues

 - #1418


ddfreyne:
I like it! (Also props for adding statistical evidence ❤️)

I’m not sure whether it makes sense to add a test — randomness is rather hard to test for…

PR #1430, opened 2019-05-02
Deep merge hrefs and resource_uris

ashmckenzie:
### Detailed description

By using `.merge` and not 'deep merging' we don't get the full list of filenames because `.merge` will allow the incoming Hash (`resource_uris_with_filenames`) to override the source Hash (`hrefs_with_filenames`).

This ultimately leads to invalid href's not being detected / reported.


da2x:
This merging no longer happens in #1422. `resource_uris_with_filenames` will only include embedded resources (only used by the mixed_content checker) and `hrefs_with_filenames` will include everything including embeds.

Could you please check that branch and verify that the links that were missed in your situation is now being detected properly?

ddfreyne:
@ashmckenzie Hey Ash! I’ve merged @da2x’s change, which I believe should fix the issue you’re seeing. Can you confirm that the issue is no longer present in `master`?

ashmckenzie:
Thanks @ddfreyne and @da2x :bow:, `master` fixes the issues I was seeing :+1:

PR #1452, opened 2019-08-14
Fix dependency tracking for sass imports

### LINKED ISSUE #1451: Improper dependency tracking for sass imports
### Steps to reproduce

1. Run `bundle exec nanoc compile`
2. Modify a single file
3. Run `bundle exec nanoc compile` again

### Expected behavior

- Goes rather fast (hopefully less than a second)
- Time taken is the sum of the time it took each "update/create/delete" steps

### Actual behavior

````
Loading site… done
Compiling site…
      update  [0.00s]  output/jobs/index.html

Site compiled in 9.31s.
```

And yes, it really took 10 seconds…

### Details

- There are 324 files in the `content` directory of this nanoc website.
- The `Loading site... done` is super quick.

Ecco:
Fixes #1451. Not sure it's the cleanest way, but it sure is the shortest way 😄

### DIFF
--- a/nanoc/lib/nanoc/filters/sass/importer.rb
+++ b/nanoc/lib/nanoc/filters/sass/importer.rb
@@ -67,8 +67,9 @@ module Nanoc::Filters::SassCommon
       @raw_filename_to_item_map[config.object_id] ||=
         {}.tap do |map|
           items.each do |item|
-            if item.raw_filename
-              path = Pathname.new(item.raw_filename).realpath.to_s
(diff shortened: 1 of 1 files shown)


ddfreyne:
This fix is correct, though is missing tests — but arguably, that is not an easy thing to test. I will give this a try after I merge this PR.

(For future readers of this PR: This fix is correct because the map that is built in-memory is rebuilt on every invocation of Nanoc, and so can't lead to outdated information that exists because information was cached.)

Thank you for the fix!

PR #1482, opened 2020-01-21
Add Regex support to @config[:prune][:exclude]

da2x:
Add Regex support in Pruner filename_excluded?. The other exclude config options all support it.


ddfreyne:
This breaks some previous behavior, which I think needs a bit more thought. E.g.:

exclude = `foo`
Now also prunes `output/foo.txt`, but previously did not.

PR #1487, opened 2020-02-22
Don't hardcode the current class in Nanoc::Checking::Check

Ecco:
### Detailed description

This very simple change avoid repeating a class name. It also has the very nice side effect of allowing one to subclass Nanoc::Checking::Check to use custom helpers.

Example with a `ensure_matches_regex` helper:

```ruby
class ContentCheck < Nanoc::Checking::Check
  def ensure_matches_regex(filename, regex)
    add_issue(name, subject: filename) if File.read(filename).match?(regex)
  end
end

ContentCheck.define(:foobar) do
  @output_filenames.each do |filename|
    ensure_matches_regex(filename, /html/)
  end
end
```

### DIFF
--- a/nanoc-checking/lib/nanoc/checking/check.rb
+++ b/nanoc-checking/lib/nanoc/checking/check.rb
@@ -18,7 +18,7 @@ module Nanoc
       attr_reader :issues
 
       def self.define(ident, &block)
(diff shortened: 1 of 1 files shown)


ddfreyne:
Nice, thank you!

Ecco:
You're welcome 😄

PR #1496, opened 2020-03-16
Ignore redirects to self (misued for cookies)

da2x:
### Detailed description

Websites increasingly common redirect URLs back onto itself to set HTTP cookies. This change just ignores redirects from /A to /A for permanent redirects. A proper fix would require to implement the `Referer` header and support session cookies. That’s too much work to support broken sites. This change could possibly ignore misconfigured websites that accidentally create permanent redirect loops without trying to force cookies upon clients.

### To do

* [ ] Tests
* [ ] Documentation
* [ ] Feature flags
* [ ] …

### Related issues


ddfreyne:
I think that approach is reasonable! I’ve pushed a trivial fix (one style issue) and merged to master.

PR #1500, opened 2020-05-02
Fixed 'nanoc()' Sass function definition

gpakosz:
### Detailed description

The `nanoc()` Sass function takes optional keyword arguments. The counterpart Ruby function must be defined with a default hash.


gpakosz:
This is happening with sass `3.4.22` and doesn't with sass `3.7.4`

PR #1505, opened 2020-06-16
nanoc-cli: Only try to clean UTF-8 chars from UTF-8 strings

### LINKED ISSUE #1502: Encoding::CompatibilityError when trying to run nanoc on Windows and Ruby 2.7.1
On a windows 10 computer with Ruby 2.7.1 and latest stable release of the nanoc gem installed (4.11.16):
Attempting to run nanoc (for example, to print the version information, or compile a new site, etc) results in an error.
I understand from the tutorial that Windows is not officially supported, but have raised this in case it is useful. I also see that the tutorial says 'Nanoc runs mostly fine on Windows, with some limitations' - are these described anywhere?

### Steps to reproduce
Here's what I did:
1. Install the gem as per the tutorial, in a cmd window
2. Type `nanoc --version` as suggested in the tutorial

### Expected behavior
Nanoc prints out its current version

### Actual behavior
Nanoc prints out some notes, warnings, and an error, as seen below (as well as a stack trace and crash log info, see underneath):
```
------------------------------------------------------------------------------
A NOTE ABOUT PERFORMANCE:
The `contracts` gem is loaded, which enables extra run-time checks, but can drastically reduce Nanoc’s performance. The `contracts` gem is intended for development purposes, and is not recommended for day-to-day Nanoc usage.

To speed up compilation, either uninstall the `contracts` gem, or use Bundler (https://bundler.io/) with a Gemfile that doesn’t include `contracts`.
------------------------------------------------------------------------------
C:/Ruby27-x64/lib/ruby/gems/2.7.0/gems/contracts-0.13.0/lib/contracts/method_reference.rb:43: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
C:/Ruby27-x64/lib/ruby/gems/2.7.0/gems/nanoc-core-4.11.16/lib/nanoc/core/dependency_props.rb:16: warning: The called method `__contracts_ruby_original_initialize_ka39pjaq9zih' is defined here

Captain! We've been hit!

Encoding::CompatibilityError: Unicode Normalization not appropriate for ASCII-8BIT
```

### Details
(Describe other details, if any, that you believe might be relevant.)
Not sure how useful it is, but note that commands such as `nanoc create-site X` appear to work - in as far as when I ran *that*, I get the following output, when creating a site called 'doc' for example:
```
------------------------------------------------------------------------------
A NOTE ABOUT PERFORMANCE:
The `contracts` gem is loaded, which enables extra run-time checks, but can drastically reduce Nanoc’s performance. The `contracts` gem is intended for development purposes, and is not recommended for day-to-day Nanoc usage.

To speed up compilation, either uninstall the `contracts` gem, or use Bundler (https://bundler.io/) with a Gemfile that doesn’t include `contracts`.
------------------------------------------------------------------------------
C:/Ruby27-x64/lib/ruby/gems/2.7.0/gems/contracts-0.13.0/lib/contracts/method_reference.rb:43: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
C:/Ruby27-x64/lib/ruby/gems/2.7.0/gems/nanoc-core-4.11.16/lib/nanoc/core/dependency_props.rb:16: warning: The called method `__contracts_ruby_original_initialize_ka39ocdx3x7d' is defined here
      create  nanoc.yaml
      create  Rules
      create  content/index.html
      create  content/stylesheet.css
      create  layouts/default.html
Created a blank Nanoc site at 'doc'. Enjoy!
```

### Crash log

(When reporting a crash, create a [Gist](https://gist.github.com/) with the contents of the `crash.log` file, and link the Gist here.)

See https://gist.github.com/ethancrawford/b6b2de0e27b31fc68c0c9e5135f955d3 for the log when running `nanoc --version`.

ethancrawford:
### Detailed description

Previously, when attempting to load a custom filter in a Windows Nanoc
installation, Nanoc was crashing with indications that the UTF-8 stream cleaner
was unable to process the supplied string.
Now, we only attempt to clean UTF-8 characters out of the string if it is UTF-8
encoded.

### To do

* [x] Tests

### Related issues

Should fix #1502. (Including the contracts gem issue).

### DIFF
--- a/nanoc-cli/lib/nanoc/cli/stream_cleaners/utf8.rb
+++ b/nanoc-cli/lib/nanoc/cli/stream_cleaners/utf8.rb
@@ -8,7 +8,7 @@ module Nanoc
       class UTF8 < Abstract
         # @see Nanoc::CLI::StreamCleaners::Abstract#clean
         def clean(str)
-          return unless str.encoding.name == 'UTF-8'
+          return str unless str.encoding.name == 'UTF-8'
 
           # FIXME: this decomposition is not generally usable
           str
--- a/nanoc-cli/lib/nanoc/cli/stream_cleaners/utf8.rb
+++ b/nanoc-cli/lib/nanoc/cli/stream_cleaners/utf8.rb
@@ -8,6 +8,8 @@ module Nanoc
       class UTF8 < Abstract
         # @see Nanoc::CLI::StreamCleaners::Abstract#clean
         def clean(str)
+          return unless str.encoding.name == 'UTF-8'
+
           # FIXME: this decomposition is not generally usable
           str
             .unicode_normalize(:nfkd)
--- a/nanoc-cli/spec/nanoc/cli/stream_cleaners/utf8_spec.rb
+++ b/nanoc-cli/spec/nanoc/cli/stream_cleaners/utf8_spec.rb
@@ -9,7 +9,7 @@ describe Nanoc::CLI::StreamCleaners::UTF8 do
     it 'does not attempt to clean the string' do
       expect(str).not_to receive(:unicode_normalize)
 
-      subject.clean(str)
+      expect(subject.clean(str)).to eq(str)
     end
   end
 
--- a/nanoc-cli/spec/nanoc/cli/stream_cleaners/utf8_spec.rb
+++ b/nanoc-cli/spec/nanoc/cli/stream_cleaners/utf8_spec.rb
@@ -3,6 +3,16 @@
 describe Nanoc::CLI::StreamCleaners::UTF8 do
   subject { described_class.new }
 
+  context 'when passed a string that is not UTF-8 encoded' do
+    let(:str) { String.new('Not UTF-8', encoding: 'ASCII-8BIT') }
+
+    it 'does not attempt to clean the string' do
+      expect(str).not_to receive(:unicode_normalize)
+
+      subject.clean(str)
+    end
+  end
+
   it 'handles all cases' do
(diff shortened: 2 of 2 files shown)


nanoc-cli/spec/nanoc/cli/stream_cleaners/utf8_spec.rb
```diff
@@ -3,6 +3,16 @@
 describe Nanoc::CLI::StreamCleaners::UTF8 do
   subject { described_class.new }
 
+  context 'when passed a string that is not UTF-8 encoded' do
+    let(:str) { String.new('Not UTF-8', encoding: 'ASCII-8BIT') }
+
+    it 'does not attempt to clean the string' do
+      expect(str).not_to receive(:unicode_normalize)
+
+      subject.clean(str)
```
ddfreyne, line 12:
```suggestion
      expect(subject.clean(str)).to eq(str)
```

nanoc-cli/lib/nanoc/cli/stream_cleaners/utf8.rb
```diff
@@ -8,6 +8,8 @@ module StreamCleaners
       class UTF8 < Abstract
         # @see Nanoc::CLI::StreamCleaners::Abstract#clean
         def clean(str)
+          return unless str.encoding.name == 'UTF-8'
```
ddfreyne, line 11: This needs to return the string itself still:

```suggestion
          return str unless str.encoding.name == 'UTF-8'
```

ethancrawford:
😅 thanks for the review 🙂

ddfreyne:
Thank you for the fix!

ethancrawford:
You're welcome 😄

PR #1508, opened 2020-08-31
Implement CoW reflink on Linux

da2x:
### Detailed description

Do a copy-on-write (CoW) filesystem-aware reflink-copying when `item_rep` is identical to the source content (`item_rep.item`). This is like a hardlink, except the two files are detached when either of them are changed. So you get the deduplication benefit, without the risk of accidentally changing a source file when modifying an output file. This assumes that `item_rep.item.filename` is a file on the disk and that it’s a source file. Fallback to a hardlink or full copy, as per the old implementation.

Supports Btrfs and XFS on Linux 4.5+.

Requires the Linux-only [reflink gem](https://codeberg.org/da/ruby-reflink). Feature is not enabled by default, but kicks in when this gem is installed.

### To do?

* [ ] Document optional gem in installation instructions.
* [ ] Existing tests cover the basic functionality. Adding tests to inspect the filesystem to determine whether a file extent is a hardlink, reflink, or regular file would be nice. However, this is a huge task and requires more specific test OS environments.

### DIFF
--- a/nanoc-core/lib/nanoc/core/item_rep_writer.rb
+++ b/nanoc-core/lib/nanoc/core/item_rep_writer.rb
@@ -62,11 +62,25 @@ module Nanoc
 
         # Write
         if is_modified
+          # Do a copy-on-write (CoW) filesystem-aware reflink-copying when item_rep is identical
+          # to the source content. Fallback to a hardlink or full copy.
+          written = false
           begin
-            FileUtils.ln(temp_path, raw_path, force: true)
-          rescue Errno::EXDEV, Errno::EACCES
-            FileUtils.cp(temp_path, raw_path)
+            require 'reflink'
+            if !content.filename.nil? && FileUtils.identical?(content.filename, temp_path)
+              Reflink.reflink_always(content.filename, raw_path)
+              written = true
+            end
+          rescue LoadError, Errno::ENOTSUP, Errno::EXDEV
           end
+          unless written
+            begin
+              FileUtils.ln(temp_path, raw_path, force: true)
+            rescue Errno::EXDEV, Errno::EACCES
+              FileUtils.cp(temp_path, raw_path)
+            end
+          end
+
         end
 
         item_rep.modified = is_modified


nanoc-core/lib/nanoc/core/item_rep_writer.rb
```diff
@@ -62,11 +62,25 @@ def write_single(item_rep, compiled_content_store, snapshot_name, raw_path, writ
...
+          # Do a copy-on-write (CoW) filesystem-aware reflink-copying when item_rep is identical
+          # to the source content. Fallback to a hardlink or full copy.
+          written = false
           begin
-            FileUtils.ln(temp_path, raw_path, force: true)
-          rescue Errno::EXDEV, Errno::EACCES
-            FileUtils.cp(temp_path, raw_path)
+            require 'reflink'
+            if !content.filename.nil? && FileUtils.identical?(content.filename, temp_path)
+              Reflink.reflink_always(content.filename, raw_path)
```
ddfreyne, line 71: I think this needs to be 

```
require 'reflink'
Reflink.reflink_always(temp_path, raw_path)
```

The `content.filename` is not necessarily set.

ddfreyne: (Though I am not sure what happens when `raw_path` already exists.)

ddfreyne: (Also, `temp_path` and `raw_path` are bad variable names. I’ll need to fix that…)

nanoc-core/lib/nanoc/core/item_rep_writer.rb
```diff
@@ -62,11 +62,25 @@ def write_single(item_rep, compiled_content_store, snapshot_name, raw_path, writ
...
         # Write
         if is_modified
+          # Do a copy-on-write (CoW) filesystem-aware reflink-copying when item_rep is identical
+          # to the source content. Fallback to a hardlink or full copy.
+          written = false
           begin
-            FileUtils.ln(temp_path, raw_path, force: true)
-          rescue Errno::EXDEV, Errno::EACCES
-            FileUtils.cp(temp_path, raw_path)
+            require 'reflink'
```
ddfreyne, line 69: If you prefer not to have the `require` here, you could probably do something like

```
begin
  require 'reflink'
rescue LoadError
end
```

and later on do `const_defined?(Reflink)`.

This is likely slightly faster, but I bet the speed difference is going to be negligible.

nanoc-core/lib/nanoc/core/item_rep_writer.rb
da2x: > The `content.filename` is not necessarily set.

I check if it is set on the line above. I specifically want to reflink the source content file (content/…) and not the working copy (temp_path). I compare them for similarity on the line above. The temp_path is deleted later (and lives on another filesystem at /tmp (tempfs)) so its completely pointless to reflink or hardlink it.

da2x:
Will open another pull with additional support for macOS.

PR #1509, opened 2020-09-10
Do copy-on-write cloning of unchanged output files.

da2x:
### Detailed description

Do a copy-on-write (CoW) filesystem-aware reflink-copy/extent cloning when item_rep is identical to the source content (item_rep.item). This is like a hardlink, except the two files are detached when either of them are changed. So you get the deduplication benefit, without the risk of accidentally changing a source file when modifying an output file. This assumes that item_rep.item.filename is a file on the disk and that it’s a source file. Fallback to a hardlink or full copy, as per the old implementation.

Requires the [clonefile gem](https://rubygems.org/gems/clonefile). Feature is not enabled by default, but kicks in when this gem is installed.

The gem supports Btrfs, OCFS2, and XFS on Linux 4.5+; and APFS on macOS 10.12.1+. (No ZFS support yet, sorry.)

### To do

* [ ] Document optional gem in installation instructions.
* [ ] Existing tests cover the basic functionality. Adding tests to inspect the files system to determine whether a file extent is a hardlink, cloned/reflink, or regular file would be nice. However, this is a huge task and requires more specific test OS environments.

### DIFF
--- a/nanoc-core/lib/nanoc/core.rb
+++ b/nanoc-core/lib/nanoc/core.rb
@@ -20,6 +20,12 @@ require 'tomlrb'
 require 'tty-platform'
 require 'zeitwerk'
 
+# Optional gems
+begin
+  require 'clonefile'
+rescue LoadError
+end
+
 module Nanoc
   module Core
     # Similar to `nil` except that it can only be compared against using
--- a/nanoc-core/lib/nanoc/core/item_rep_writer.rb
+++ b/nanoc-core/lib/nanoc/core/item_rep_writer.rb
@@ -61,11 +61,23 @@ module Nanoc
         end
 
         # Write
+        # Clone or copy the file on a copy-on-write file system (APFS, Btrfs, XFS) if the
+        # resulting and source files are identical.
+        # Otherwise, hardlink the file or do a regular copy if unsupported.
         if is_modified
-          begin
-            FileUtils.ln(temp_path, raw_path, force: true)
-          rescue Errno::EXDEV, Errno::EACCES
-            FileUtils.cp(temp_path, raw_path)
+          written = false
+          if Object.const_defined?('Clonefile')
+            if !content.filename.nil? && FileUtils.identical?(content.filename, temp_path)
+              Clonefile.auto(content.filename, raw_path)
+              written = true
+            end
+          end
+          unless written
+            begin
+              FileUtils.ln(temp_path, raw_path, force: true)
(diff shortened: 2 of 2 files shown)


da2x:
The failing test appears to be timing based. I don’t have permissions to restart it. (Shouldn’t those sleepy tests be using fsync instead?)

ddfreyne:
I need some more effort to think about this solution, because I think this might not be the best way to implement it — a copy is stored in `tmp` anyway, and so a hardlink from there is fine (unless you manually edit stuff in tmp/, which you shouldn’t do anyway.)

If the problem we’re trying to solve is the disk space overhead of tmp/, then it makes the most sense to have a COW close from content/ into tmp/ and into output/, which this PR doesn’t do, and so it doesn’t solve the disk space issue.

Unfortunately, letting Nanoc support COW clones from content/ into tmp/ and output/ isn’t straightforward and needs more thought.

da2x:
@ddfreyne please clarify what you mean by tmp? nanoc-dir/tmp? or /tmp/? You can’t link or clone across mount point boundaries. On all (?) major Linux distributions, /tmp is either a tmpfs partition or a separate file system partition on low-memory setups. So you always need to do a full copy into /tmp anyway.

> If the problem we’re trying to solve is the disk space overhead of tmp/, then it makes the most sense to have a COW close from content/ into tmp/ and into output/, which this PR doesn’t do, and so it doesn’t solve the disk space issue.

The problem I’m trying to solve is to break the hardlinks from content/ to output/ in a file-system-space-efficient way. Modifying content/ changes the output/ and visa versa. This is the current behaviour on master.

This change does implement cloning from content/ (the content.filename variable) to output/ (the raw_path variable). The temp_path variable holds either a reference to /tmp (if modified) or the original file in content/ (unless modified).

da2x:
I edited and clarified the previous comment after posting it, so please don’t read the slightly less concise email notification summary. 😆

ddfreyne:
Oh yes — I think my previous comment had some outdated assumptions about Nanoc in mind (the disadvantage of maintaining software for 10+ years is that it changes so much and you keep all the individual versions of how it used to work in mind).

I’m going to give a shot at completing this — it might be simpler than I stated!

ddfreyne:
Here’s my own take on clonefile, based off your code: #1511. I’ve tested this out locally and it works like a charm (build fails because I messed up my `git` setup for now).

da2x:
Replaced by #1511. :+1:

PR #1535, opened 2021-01-01
Fix GitHub Actions bundle install path

opatry:
### Detailed description

Using `bundle install --path` is deprecated, use recommended/up to date way to configure install path.

```
[DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set --local path 'vendor/bundle'`, and stop using this flag
```

I also took the opportunity to install gems in parallel.

### Related issues

Follow up of nanoc/nanoc/pull/1534

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -23,7 +23,9 @@ jobs:
           path: vendor/bundle
           key: bundler-b-${{ runner.os }}-${{ matrix.ruby-version }}-${{ hashFiles('Gemfile') }}
 
-      - run: bundle install --path=vendor/bundle
+      - run: |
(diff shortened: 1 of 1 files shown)


opatry:
See details here: https://bundler.io/v2.2/man/bundle-config.1.html#REMEMBERING-OPTIONS

ddfreyne:
Perfect, thank you!

ddfreyne:
Something still to figure out is why sometimes the tests either hang, or sometimes segfault. No luck figuring that out so far.

opatry:
Yes I saw that, no clue 🤷‍♂️.

Always with Ruby 2.7 apparently...

Olivier Patry

PR #1536, opened 2021-01-01
Debug why nanoc tests sometimes time out

ddfreyne:
As it says on the tin.

**Not intended to be merged.**


ddfreyne:
It appears that I cannot reproduce the freezing/timeout anymore, although the segfault issue remains.

PR #1540, opened 2021-02-27
Spelling and more descriptive error

da2x:
I've seen this randomly for as long as I've used Nanoc, but I've never bothered to read the code to learn that it was caused by my use of `raw_path`. I thought it was just some random item that sometimes randomly failed. (E.g. sitemaps cause this through `File.mtime(item.raw_path)`.)

I can calculate the item[:mtime] in the preprocessor and save myself random errors, but first I needed to learn that this was caused by `raw_path`. Hoping the new error can help others better understand what happens and save them time.


ddfreyne:
Thank you!

PR #1557, opened 2021-11-29
Log the filename when it is not absolute

jwillemsen:
### Detailed description

Log the filename when it is not absolute to give better feedback to the user


ddfreyne:
Thanks! I merged this manually (see cec6a32721a31dacf88b58cb7a43445952cedaa6) and increased test coverage for it (5714132c15af403184e895466d094b355de3e910).

Additionally, future PRs will properly have their tests run with GitHub actions — this was an oversight when I set up CI on GitHub.

PR #1560, opened 2022-01-11
add Date to the list of permitted classes for YAML

dmathieu:
### Detailed description

In the latest versions of Psych (such as the one running with Ruby 3.1), only the Symbol class is allowed to be parsed.
See https://github.com/ruby/psych/blob/master/lib/psych.rb#L369

This change adds the `Date` class to the allowed objects. It's pretty common for date objects to be within metadata (for a `created_at` attribute in a blog for example).

### To do

* [x] Tests

### DIFF
--- a/nanoc/lib/nanoc/data_sources/filesystem/parser.rb
+++ b/nanoc/lib/nanoc/data_sources/filesystem/parser.rb
@@ -4,6 +4,7 @@
 class Nanoc::DataSources::Filesystem
   class Parser
     SEPARATOR = /(-{5}|-{3})/.source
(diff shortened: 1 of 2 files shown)


ddfreyne:
Thanks, looks good!

I have merged this manually (see b3755eca87b4372c8755dfc170f38653fb15a710) and fixed new style issues (see c6fbcd7e208158804fb102753cc2be97f9af7399).

I’ve also set up GitHub CI to work on PRs (I missed that in the initial setup) and will set up GitHub CI to also run Rubocop, so future PRs will get proper feedback!

PR #1574, opened 2022-04-26
Allow same Yaml classes for binary and frontmatter

da2x:
Psych-4.0 changes to safe_load by default. Compilation breaks with below error for anyone with binary files accompanied by a Yaml file with a date. 

```
Psych::DisallowedClass: Tried to load unspecified class: Date
```

The DateClass isn’t allowed in safe_load mode. Nanoc expects it to be for e.g. created_at. Duplicating code from parser.b to allow same classes in all Yaml source files and front matter.


da2x:
I doubt that test failure is related to this change.

ddfreyne:
Thanks! This change makes sense.

(The automated GitHub tests are flaky unfortunately… will need to look into that once more.)

PR #1579, opened 2022-06-04
Replace UglifyJS with Terser

### LINKED ISSUE #1578: replace uglifier with terser gem
### Steps to reproduce

terser is better maintained fork of uglify-js and most projects have switched to terser. terser gem provides similar functionality to replace uglifier gem.

### Expected behavior

Should use terser instead of uglifier.

### Actual behavior

Still using uglifier

### Details

In debian we are trying to remove uglifier and nanoc is the last dependency blocking it. See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=981224

ddfreyne:
### Detailed description

UglifyJS is unmaintained. This PR replaces the `uglify_js` filter with a `terser` filter, which does essentially the same.

The new filter should be backwards compatible. The old filter identifier, `uglify_js`, is kept for that reason.

Additionally, this PR has some development improvements regarding the JavaScript implementation. This PR switches the JavaScript implementation to use `mini_racer` rather than `libv8`, which makes it work on macOS by default. It also makes the `duktape` gem obsolete.

### To do

* [x] Tests
* [x] Verify backwards compatibility

### Related issues

Fixes #1578.


Gemfile
```diff
@@ -58,11 +57,11 @@ group :plugins do
   gem 'haml'
   gem 'kramdown'
   gem 'less', '~> 2.6', platforms: :ruby
-  gem 'libv8', platforms: :ruby
```
opatry, line 61: @ddfreyne it seems it breaks CI build

```
Errno::ENOENT: No such file or directory @ apply2files - /home/runner/work/nanoc/nanoc/vendor/bundle/ruby/2.6.0/gems/libv8-node-16.10.0.0-x86_64-linux/vendor/v8/x86_64-linux-musl/libv8/obj/libv8_monolith.a
```

opatry:
@ddfreyne I incorrectly assumed such update was available in last `nanoc` release.
I checked too quickly when updating my gems, saw an update of `nanoc` and considered it was the one I wanted 😅.

Do you have an ETA for the next release to include the `:terser` filter availability?
(Obviously, no pressure at all!)

Gemfile
ddfreyne: Yeah — apparently on Ruby 2.6 only though, which is EOL so I plan on dropping Ruby 2.6 support rather than fixing this.

ddfreyne:
@opatry I’ll get a release out today!

ddfreyne:
Done. This is released in 4.12.7.

PR #1589, opened 2022-09-01
Escape URLs so terminals can detect them properly

da2x:
### Detailed description

Escape links as `<url>`, so terminals don’t interpret unwanted characters as being part of the link. E.g. the link from the message `broken reference to https://nanoc.app/: error` would be interpreted as `https://nanoc.app/:` instead of `https://nanoc.app/` as expected. This small change fixes that issue.

### To do

All done and I’ve updated the affected tests.


da2x:
The two failing tests are present on the parent commit and were not introduced by this pull request.

denisdefreyne:
Looks good!

I’ll need to spend some time on fixing build failures on the `main` branch -- some new dependency versions have broken things more than I’d like.

PR #1598, opened 2022-10-01
Performance: Speed up NotWritten outdatedness rule

denisdefreyne:
### Detailed description

This changes the `NotWritten` to pre-load the list of files in the output directory, so that the filesystem only has to be accessed once, rather than for every file.

This leads to a nice speedup. For one particular site:

```
                                          x │ count     min     .50     .90     .95     max     tot
────────────────────────────────────────────┼──────────────────────────────────────────────────────
 Nanoc::Core::OutdatednessRules::NotWritten │  3704   0.00s   0.00s   0.01s   0.01s   0.05s   4.77s
```

```
                                          x │ count     min     .50     .90     .95     max     tot
────────────────────────────────────────────┼──────────────────────────────────────────────────────
 Nanoc::Core::OutdatednessRules::NotWritten │  3704   0.00s   0.00s   0.00s   0.00s   0.05s   0.07s
```

Before: 4.77s. After: 0.07s. That’s a 60x–70x speedup of the `NotWritten` rule.

### To do

* [x] Tests
* [x] Check in output directory, not in `Dir.pwd`

Future work could entail excluding directories such as `.git/`, but that’d likely require switching from `Dir.glob` to `Find`, which might not yield much of a speedup. In any case, the `NotWritten` rule is no longer a bottleneck now.

### Related issues

n/a

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -35,6 +35,7 @@ Naming/MethodParameterName:
   AllowedNames:
     - "to" # destination
     - "e" # element
+    - "fn" # filename
     - "id" # identifier
     - "io" # IO instance
     - "y" # enumerator yielder
--- a/nanoc-core/lib/nanoc/core/outdatedness_rules/not_written.rb
+++ b/nanoc-core/lib/nanoc/core/outdatedness_rules/not_written.rb
@@ -6,10 +6,30 @@ module Nanoc
       class NotWritten < Nanoc::Core::OutdatednessRule
         affects_props :raw_content, :attributes, :compiled_content, :path
 
-        def apply(obj, _outdatedness_checker)
-          if obj.raw_paths.values.flatten.compact.any? { |fn| !File.file?(fn) }
+        def apply(obj, outdatedness_checker)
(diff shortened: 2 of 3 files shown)


opatry:
Sounds nice!

I'm quite interested by this kind of improvements given my personal photoblog being quite heavy in terms of files in output.

Can you elaborate a bit on use cases where your change can have impact?

Eg on my side:
- 3554 generated file(s)
  - 550 html
  - 1487 jpg
  - 1487 webp
  - other

denisdefreyne:
@opatry Interestingly, now that I’m re-running the performance tests for this change, I see a much smaller benefit.

I had some trouble with my MacBook Air M2 being throttled, presumably because it ran quite hot.

I don’t get results with `File.file?` that are even *close* to 4.77s anymore. Old `File.file?` approach: ~0.04s. New `Dir.glob` approach: ~0.02s.

I’ll have to do some more thorough testing, but this might have not had nearly as much of an impact as I had hoped.

opatry:
Still, better than nothing 😜.

PR #1604, opened 2022-10-15
Improve performance (part III)

denisdefreyne:
### Detailed description

This continues the progress made in improving performance.

The main learning: `Set#any?` is *slow* -- much better to use `Set#empty?` and negate.

I also realised that `Set`s can be avoided if the possible elements are limited and known ahead of time. `Set` is rather slow, and more verbose code can be a lot faster.

### To do

* [ ] Tests
  * It’d be nice to have a test case for `#props_for` returning nil, but by now it’s less and less likely that there’s such old versions of Nanoc around that require this very tiny bit of backwards compatibility.

### Related issues

Two other performance-focused PRs:
* #1602 
* #1603

### DIFF
--- a/nanoc-cli/spec/nanoc/cli/commands/show_data_spec.rb
+++ b/nanoc-cli/spec/nanoc/cli/commands/show_data_spec.rb
@@ -55,42 +55,6 @@ describe Nanoc::CLI::Commands::ShowData, stdio: true do
       end
     end
 
-    context 'dependency (without props) from config to dog' do
-      before do
-        dependency_store.record_dependency(item_dog, config)
-      end
-
-      it 'outputs no dependencies for /about.md' do
-        expect { subject }.to output(%r{^item /about.md depends on:\n  \(nothing\)$}m).to_stdout
-      end
-
-      it 'outputs dependencies for /dog.md' do
-        expect { subject }.to output(%r{^item /dog.md depends on:\n  \[  config \] \(racp\)$}m).to_stdout
-      end
-
-      it 'outputs no dependencies for /other.dat' do
-        expect { subject }.to output(%r{^item /other.dat depends on:\n  \(nothing\)$}m).to_stdout
-      end
-    end
-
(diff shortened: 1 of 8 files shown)


DivineDominion:
@denisdefreyne I'm curious: do you have a hunch why `Set#any?` is slow? Once you have everything hashed, which might take a while sure, the hash lookup should be super quick.

denisdefreyne:
`Set#any?` is slow because `Set` does not implement `#any?` directly, but delegates to `Enumerable#any?` instead. `Enumerable#any?` seems to be implemented in a way in which it starts iterating over elements, and then breaks out after the first element, something like this:

```ruby
module MyEnumerable
  def any?
    each do
      return true
    end
  
    false
  end
end
```

That is what the CPU profile seems to suggest, at least.

It also explains the significance difference in performance. Here’s a benchmark:

```
Warming up --------------------------------------
                any?   276.965k i/100ms
             !empty?     3.323M i/100ms
Calculating -------------------------------------
                any?      2.756M (± 2.6%) i/s -     13.848M in   5.027698s
             !empty?     33.343M (± 0.2%) i/s -    169.469M in   5.082677s

Comparison:
             !empty?: 33342764.7 i/s
                any?:  2756310.9 i/s - 12.10x  (± 0.00) slower
```

Benchmark implementation:

```ruby
require 'benchmark/ips'
require 'set'

set = Set.new
100.times { set << rand }

Benchmark.ips do |x|
  x.report("any?") do |times|
    i = 0
    while i < times
      set.any?
      i += 1
    end
  end

  x.report("!empty?") do |times|
    i = 0
    while i < times
      !set.empty?
      i += 1
    end
  end

  x.compare!
end
```

denisdefreyne:
One more thing: `#any?` also returns false for non-empty collections that contain falsy elements: `[false, nil].any?` is `false`.

So, switching from `xyz.any?` to `!xyz.empty?` is not only faster, but also more correct. (Though in Nanoc’s case, the correctness wasn’t an issue.)

PR #1629, opened 2022-11-26
Write dart sass importer in idiomatic ruby

ntkme:
This PR rewrites the dart sass custom importer in idiomatic ruby code.


denisdefreyne:
Thanks! This refactoring totally makes sense.

PR #1643, opened 2023-01-20
Add org-mode filter

antaz:
### Detailed description

Added support for emacs' markup language [org-mode](https://orgmode.org/) through the `:orgmode` filter

### To do

* [x] Partial testing
* [x] Documentation
* [x] Feature flags


blocr:
@denisdefreyne any chance for this to be considered?

denisdefreyne:
Hey there, thanks for the contribution! I’m short on time to look at this properly yet, but this looks like valuable addition to Nanoc.

denisdefreyne:
Apologies for taking a long time with this. It’s still on my list of things to get to.

One thing I’ll change is move the `org-mode` filter into its own gem, so it can be installed separately. This would be similar to how the newer gems, `nanoc-tilt` and `nanoc-dart-sass`, are implemented.

(If you want, you can take on this work as well — it might be faster — but I’ll get to it at some point, for sure.)

denisdefreyne:
A new PR that uses the separate-gem approach is here: #1688

denisdefreyne:
This is now released in the `nanoc-org-mode` gem. To use the new gem, add `nanoc-org-mode` to the `nanoc` group of your Gemfile:

```ruby
group :nanoc do
  gem 'nanoc-org-mode'
end
```

… and then call the `:org_mode` filter:

```ruby
compile '/**/*.org' do
  filter :org_mode
  layout '/default.*
end
```

Thank you for your contribution @antaz! (And thank you for your patience!)

PR #1645, opened 2023-02-26
Add support for `filter [:binary, :text] => ...`

0x1eef:
### Detailed description

Hello :wave: 
This change adds support for creating filters that can come from a binary, or textual item. 
It would be a nice-to-have for [0x1eef/nanoc-gunzip.rb](https://github.com/0x1eef/nanoc-gunzip.rb).

Example:

```ruby
class MyFilter < Nanoc::Filter
  identifier :my_filter
  type [:binary, :text] => :text

  def run(content_or_filename, options = {})
    if @item.binary?
      # Handle binary item
    else
      # Handle textual item
    end
  end
end
```
### To do

* [ ] Tests
* [ ] Documentation
* [ ] Feature flags
* [ ] …

### Related issues

### DIFF
--- a/nanoc-core/lib/nanoc/core/executor.rb
+++ b/nanoc-core/lib/nanoc/core/executor.rb
@@ -107,7 +107,9 @@ module Nanoc
         klass = Nanoc::Core::Filter.named!(filter_name)
 
         last = @compilation_context.compiled_content_store.get_current(@rep)
-        if klass.from_binary? && !last.binary?
+        if klass.from_binary_or_text?
+          # do nothing
+        elsif klass.from_binary? && !last.binary?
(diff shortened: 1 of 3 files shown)


0x1eef:
This is how nanoc-gunzip.rb looks with this change: 
https://github.com/0x1eef/nanoc-gunzip.rb/commit/4726cb4c597d06de6b75abba8acf437c73c22483

I tested it with one of my sites, and it seems to work. I'm having a little trouble installing all of nanoc's Gemfile dependencies locally. I was able to confirm `executor_spec.rb` is green, but I'm not sure if all the test suite is green or not.

0x1eef:
It does not look like there's interest in this change. I will close the PR but send a ping if I'm mistaken.

PR #1649, opened 2023-06-01
Remove redundant defaults in command descriptions

denisdefreyne:
### Detailed description

When a `:default` is specified, it is automatically added to the command description, so it does not need to be specified in the command description manually. In fact, adding it to the command description manually makes it appear twice:

```
    -o --host=<value>         specify the host to listen on (default:
                              127.0.0.1) (default: 127.0.0.1)
```

### To do

* [x] Tests

### Related issues

n/a


denisdefreyne:
Style issues are making the build fail -- will fix these in a future PR.

PR #1650, opened 2023-06-01
Disable RDiscount on clang 16

denisdefreyne:
### Detailed description

RDiscount does not compile with clang 16 (see https://github.com/davidfstr/rdiscount/issues/155). This disables RDiscount for now.

### To do

n/a

### Related issues

n/a


denisdefreyne:
Style issues are making the build fail -- will fix these in a future PR.

PR #1654, opened 2023-06-13
Adapt tests to rouge 4

terceiro:
The exact contents produced by the GitHub theme in rouge 4 has been updated and this PR adapts the test to pass against rouge 4 (it will fail against rouge < 4 given the output is different).


denisdefreyne:
Thanks for the patch!

I think it’s fine to not officially support Rouge 3.x anymore, so sticking to Rouge 4.x is fine (otherwise we’d have to reintroduce https://github.com/thoughtbot/appraisal).

PR #1657, opened 2023-06-18
Make error handler print #full_message if available

denisdefreyne:
### Detailed description

If an error implements `#full_message`, it will be printed after the error detail. This makes the `nanoc-dart-sass` errors much more interesting:

```
Captain! We’ve been hit!

Sass::CompileError: Undefined mixin.

Error: Undefined mixin.
  ╷
9 │   @include oops;
  │   ^^^^^^^^^^^^^
  ╵
  nanoc:/assets/style/_2021q4_common.* 9:3  @use
  - 7:1                                     root stylesheet

Current item: /assets/style/2021q4.scss (:default representation)
```

### To do

* [x] Tests

### Related issues

n/a

### DIFF
--- a/nanoc-cli/lib/nanoc/cli/error_handler.rb
+++ b/nanoc-cli/lib/nanoc/cli/error_handler.rb
@@ -115,6 +115,7 @@ module Nanoc::CLI
       stream.puts 'Captain! We’ve been hit!'
 
       write_error_message(stream, error)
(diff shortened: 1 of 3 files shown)


DivineDominion:
Tested in my project, looks great 👍

denisdefreyne:
This is released in 4.12.16 now.

PR #1664, opened 2023-09-20
Update for breaking change in sass-embedded 1.68.0

ntkme:
### Detailed description

In sass-embedded 1.68.0, there is a breaking change in importer function signature.

See: https://github.com/ntkme/sass-embedded-host-ruby/pull/156

### DIFF
--- a/nanoc-dart-sass/lib/nanoc/dart_sass/filter.rb
+++ b/nanoc-dart-sass/lib/nanoc/dart_sass/filter.rb
@@ -31,7 +31,7 @@ module Nanoc
           @source_item = source_item
         end
 
-        def canonicalize(url, **)
+        def canonicalize(url, *, **)
           # Construct proper URL with `nanoc:` prefix if needed
           if url.start_with?('nanoc:')
             url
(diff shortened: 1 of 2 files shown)


nanoc-dart-sass/lib/nanoc/dart_sass/filter.rb
```diff
@@ -31,7 +31,7 @@ def initialize(items, source_item)
           @source_item = source_item
         end
 
-        def canonicalize(url, **)
+        def canonicalize(url, *, **)
```
ntkme, line 34: - `sass-embedded < 1.68.0`: `def canonicalize(url, from_import:)`
- `sass-embedded >= 1.68.0`: `def canonicalize(url, canonicalize_context)`

Since we don't use the context, we can define this method as `def canonicalize(url, *, **)`, which is compatible with both all past sass-embedded versions.

ntkme:
@denisdefreyne Sorry about the inconvenience of having a breaking change. Please take a look when you get a chance. Thanks!

denisdefreyne:
Thank you for the fix!

PR #1665, opened 2023-10-10
fix: wrong extension for non-index .md files

henk84:
### Detailed description

The initial rules file contains an example for handling `.md` files but doesn’t set the correct extension for the resulting `.html` file. This change strips the `.md` extension and adds a `.html` extension.

### To do

* Tests


denisdefreyne:
Excellent catch!

PR #1669, opened 2023-11-30
Set metadata source_code_uri

sato11:
This commit adds 'source_code_uri' key to gemspec's metadata, which makes nanoc and the plugins compatible with  [gem-src](https://rubygems.org/gems/gem-src)'s functionality: without it there is no way to detect the source repo unless actually opening the homepage and navigate through it, which I believe is rather tedious.

The 'source_code_uri' item in gemspec's metadata is described in [the guideline rubygems.org publishes](https://guides.rubygems.org/specification-reference/#metadata), and it is [what gem-src uses to prioritize when inferring which repo to clone](https://github.com/amatsuda/gem-src/blob/v0.10.2/lib/rubygems_plugin.rb#L38).

I guess there are not many who have come across this issue, but what do you think of it anyway?

And last but not least, thanks for letting me have a chance to touch your product. I have had a lot of fun and I appreciate your effort 🤝

### DIFF
--- a/guard-nanoc/guard-nanoc.gemspec
+++ b/guard-nanoc/guard-nanoc.gemspec
@@ -22,5 +22,8 @@ Gem::Specification.new do |s|
 
   s.files         = Dir['[A-Z]*'] + Dir['lib/**/*'] + ['guard-nanoc.gemspec']
   s.require_paths = ['lib']
-  s.metadata['rubygems_mfa_required'] = 'true'
+  s.metadata = {
+    'rubygems_mfa_required' => 'true',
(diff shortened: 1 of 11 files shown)


denisdefreyne:
@sato11 Thank you!

I made two changes:

1. Fix Rubocop style (`rubocop -a`)

2. Fix the URL for `nanoc`. This one is tagged with *just* the version number, not as `nanoc-v#{VERSION}` for historical reasons.

In any case, the change is merged!

denisdefreyne:
This is released in Nanoc 4.12.19.

sato11:
@denisdefreyne Thanks for your passionate work!

PR #1675, opened 2023-11-30
Allow slim 5.x

denisdefreyne:
### Detailed description

This allows slim 5.x.

### To do

n/a

### Related issues

n/a

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -76,7 +76,7 @@ group :plugins do
   gem 'ruby-handlebars'
   gem 'rubypants'
   gem 'sass'
-  gem 'slim', '~> 4.0'
+  gem 'slim', '~> 5.0'
   gem 'terser'
   gem 'typogruby'
   gem 'w3c_validators'


denisdefreyne:
@zor-el The `Gemfile` specifies the version that is used for development/testing; Nanoc still works with Slim 4.x (but Nanoc is only tested with Slim 5.x rather than 4.x now).

In hindsight, the PR title “Allow slim 5.x” is wrong regardless: it should’ve said “Run tests with Slim 5.x rather than 4.x”

zor-el:
Thanks @denisdefreyne, I'm aware that the Gemfile only affects contributors. My concern was that this meant support for Slim <4 is being phased out. But if I understood you correctly, it doesn't. 👍

zor-el:
BTW, I've been very impressed by the mostly backward compatibility preserving development of this project, which is why I find myself coming back to it over and over.

zor-el:
> Generally, only the latest version of dependencies (such as Slim etc) is tested against. It is possible to run against more than one version (e.g. with [the appraisals gem](https://github.com/thoughtbot/appraisal)) but it’s not particularly straightforward to set up.

Just out of curiosity, is the reason why Gemfile.lock isn't under version control as Bundler recommends connected to this? 🤔

PR #1679, opened 2023-12-26
Specify pry as one of nanoc-cli's runtime dependencies

### LINKED ISSUE #1678: Unable to run `nanoc shell` without explicitly adding `gem 'pry'` to the application's Gemfile
### Steps to reproduce
Run `nanoc shell`, without `pry` in the Gemfile.

### Expected behavior
It should start the pry session.

### Actual behavior
It stops instantly with this message:

```
Error: cannot load such file -- pry

1. Add `gem 'pry'` to your Gemfile
2. Run `bundle install`
3. Re-run this command
```

### Details
We can just add `gem 'pry'` as instructed, but doesn't it feel strange, that we are forced to bring in an additional dependency to use the library that we have already installed through the Gemfile? I would want `nanoc shell` to run without being asked to modify anything in the application's side.

### Crash log
N/A

sato11:
### Detailed description
This PR adds a runtime dependency on pry to nanoc-cli so it does not fail when the application's Gemfile does not have pry.

I suppose that in the relatively old days pry was so prevalent that this was not thought of as a problem. Though nowadays irb is getting popularized again which sometimes displaces pry, when it becomes a problem.

### To do
None. 

### Related issues
Closes #1678.

### DIFF
--- a/nanoc-cli/nanoc-cli.gemspec
+++ b/nanoc-cli/nanoc-cli.gemspec
@@ -20,6 +20,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('cri', '~> 2.15')
   s.add_runtime_dependency('diff-lcs', '~> 1.3')
   s.add_runtime_dependency('nanoc-core', "= #{Nanoc::CLI::VERSION}")
+  s.add_runtime_dependency('pry', '>= 0')
   s.add_runtime_dependency('zeitwerk', '~> 2.1')
   s.metadata = {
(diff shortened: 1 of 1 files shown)


sato11:
I guess the CI failing to build mini_racer is a separate complication which has something to do with cache. I'll chase it also although it does not seem to become clear easily 🙁

sato11:
I have reasoned out the problem at #1680. I'll keep it separate so will you give it a look before this one?

denisdefreyne:
Thank you!

PR #1691, opened 2024-04-01
nanoc-dart-sass: somewhat hacky support for partials.

bobbens:
### Detailed description

Current nanoc-dart-sass doesn't have support for partials (files starting with '_'). This is a quite ugly fix, but it does work. Feel free to close if you have a better solution or find it too ugly to merge.

### To do

* [x] Support for partials

### Related issues

#1690

Issue #1690, opened 2024-04-01
dart_sass incompatibility with sass with @import and partials (_ prefix)


bobbens:
### Steps to reproduce

1. Set up repository with bootstrap source code.
2. Set up Rule to compile bootstrap source code with dart_sass.
3. Run Nanoc.

### Expected behavior

Clean compilation.

### Actual behavior

Errors on all the `@import` which are trying to include a partial (file starts with `_`).

### Details

The core of the issue is that the old `sass` plugin will cleanly include all these modules that start with `_`, however, the dart_sass one does not, and gives an error on each one. Of course, it should be possible to prepend `_` to all the includes (or add a filter to do that), but for drop-in compatibility, it would make sense to have the same behaviour as sass, especially as this is the latest bootstrap.

```
Sass::CompileError: /home/ess/naev/docs/lore/vendor/bundle/ruby/3.0.0/gems/nanoc-dart-sass-1.0.2/lib/nanoc/dart_sass/filter.rb:74:in `find_item_
for_url': Could not find an item matching pattern `/scss/mixins/banner.*` (RuntimeError)
...

1 │ @import "mixins/banner";
  │         ^^^^^^^^^^^^^^^
  ╵
  nanoc:scss/_bootstrap 1:9  @import
  - 1:9                      root stylesheet
  ...
  ```
  
Note that the file `/scss/mixins/_banner.scss` does exist, but `/scss/mixins/banner.scss` does not.

The same repo cleanly compiles with the `sass` filter. Only does it fail with `dart_sass`.

Reference for partials: https://sass-lang.com/documentation/at-rules/import/#partials

### Crash log

https://gist.github.com/bobbens/017370417a6c1187cb311d26f958702b

**Comment by bobbens** [comment]
The following horrible modification to `vendor/bundle/ruby/3.0.0/gems/nanoc-dart-sass-1.0.2/lib/nanoc/dart_sass/filter.rb` at around line 73  'fixes' it:

```
          unless item
            pat = File.dirname(pat)+"/_"+File.basename(pat)
            item = @items[pat]
            unless item
              raise "Could not find an item matching pattern `#{pat}`"
            end
          end
```

denisdefreyne:
Fixed in #1699. Release coming soon.

denisdefreyne:
This is released in `nanoc-dart-sass` v1.0.3.

PR #1699, opened 2024-04-19
nanoc-dart-sass: Support partial and directory imports

denisdefreyne:
### Detailed description

This makes nanoc-dart-sass support importing partials (e.g. `@import 'foundation';` loads `_foundation.scss`) and directory imports (e.g. `@import 'foundation';` loads `foundation/_index.scss`).

It also handles ambiguous better: if multiple items match, an error is raised.

It was tough to reuse [the existing `sass-embedded-shim-ruby` implementation](https://github.com/sass-contrib/sassc-embedded-shim-ruby/blob/72399fda8dcc3eb511d7914fedf6e9d4a705d073/lib/sassc/embedded.rb#L226-L295) because nanoc-dart-sass allows Nanoc-style patterns in its imports, which regular dart-sass doesn’t do. It’s both a drawback and an advantage.

### To do

* [x] Tests

### Related issues

- #1690
- #1691

### DIFF
--- a/nanoc-dart-sass/lib/nanoc/dart_sass/filter.rb
+++ b/nanoc-dart-sass/lib/nanoc/dart_sass/filter.rb
@@ -68,13 +68,24 @@ module Nanoc
               File.expand_path(pat, dirname)
             end
 
-          item = @items[pat]
+          items = []
 
-          unless item
+          # Try as a regular path
+          items << @items[pat]
+
+          # Try as a partial
+          partial_pat = File.join(File.dirname(pat), "_#{File.basename(pat)}")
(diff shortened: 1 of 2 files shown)


denisdefreyne:
@bobbens Could you give this a try to see how this works for you? It would replace #1691.

FYI @ntkme I found it difficult to reuse the existing implementation (which you linked to in #1691 -- thank you!) because nanoc-dart-sass supports Nanoc patterns in imports. Whether or not this is a good thing is up for debate 🙃

bobbens:
I tested it and it seems to work fine. Thanks!

PR #1716, opened 2024-11-02
Strip drive identifier from File.expand_path on Windows

### LINKED ISSUE #1703: dart-sass filter does not handle imports correctly on Windows
### Steps to reproduce

1.  have a .scss file that imports another .scss file with an `@import` line, e.g.  `@import 'components/colors';`
2. compile on Windows

### Expected behavior

It should compile.

### Actual behavior

It compiles fine on Mac but on Windows it shows:
Error "Could not find an item matching pattern `C:/components/color`"

### Details

The problem seem to be [this line](https://github.com/nanoc/nanoc/blob/0ca9f97bd6ee2a1e46c610cf44449cca6c1b7831/nanoc-dart-sass/lib/nanoc/dart_sass/filter.rb#L65)
`File.expand_path(pat, dirname)` adds a drive identifier on Windows. If I patch that line to read `File.expand_path(pat, dirname).sub(/^[A-Z]:/,'')` then it works.

Somewhat related note: if I pass a custom Sass function to the filter through a Proc then it seems to throw off the outdatedness checker and always recompile the entire site, but only if the environment is not "default". But that probably does not have anything to do with this.

denisdefreyne:
### Detailed description

On Windows, `File.expand_path` adds a drive identifier to the beginning (e.g. `C:`) which breaks things when a Nanoc path is expected (one that starts with a slash).

The fix isn’t particulary pretty, but I cannot come up with a better solution.

:warning: I also was not able to test this on Windows yet.

### To do

* [x] Tests

### Related issues

Fixes #1703.

### DIFF
--- a/nanoc-core/lib/nanoc/core/utils.rb
+++ b/nanoc-core/lib/nanoc/core/utils.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+# .sub(/^[A-Z]:/,'')
+
+module Nanoc
+  module Core
+    # Utilities that don’t fit anywhere else.
+    #
+    # @api private
+    module Utils
+      # Same as File.expand_path, but does not add a drive identifier on
+      # Windows. This is necessary in case the path is a Nanoc path, rather than
+      # a filesystem path.
+      def self.expand_path_without_drive_identifier(file_name, dir_string)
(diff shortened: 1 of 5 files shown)


denisdefreyne:
Verified on Windows!

(Though the tests on Windows really need some work… many are failing due to platform differences.)

Issue #1717, opened 2024-11-05
Use feature detection instead of platform detection if possible


Fjan:
The dart-sass filter is currently non-functional on some PCs due to a problem with the tty-platform gem that we use to detect whether we are on Windows. (https://github.com/nanoc/nanoc/issues/1703, https://github.com/piotrmurach/tty-platform/issues/1) 
While this will probably resolve itself in time, it's a bit brittle to rely on a platform string that might break as new versions of ruby are released. Instead I think it would be better to feature detect that we are on a Windows file system, like this:

```ruby
    def windows_fs?
      ::File::ALT_SEPARATOR == "\\"
    end
```

At the moment the dart-sass filter seems to be the only place in nanoc we use this windows detection, I think? For detecting other windows related properties we may need to add additional helper functions, but in general I think this would be preferable to depending on a platform to be reliably detected and to infer functionality from that.

denisdefreyne:
There is a possible fix in #1718. Can you check whether this works for you?

Fjan:
Thanks, looks good to me. I don't have access to a PC with git on it though, so can only verify it after release I'm afraid.

PR #1718, opened 2024-11-12
Improve Windows file path detection

### LINKED ISSUE #1703: dart-sass filter does not handle imports correctly on Windows

denisdefreyne:
### Detailed description

This makes the detection of Windows file paths (starting with a drive identifier, e.g. `C:\blah`) more reliable.

### To do

Tests? (This is hard to test automatically though.)

### Related issues

Fixes #1703, #1717.

### DIFF
--- a/nanoc-core/lib/nanoc/core/utils.rb
+++ b/nanoc-core/lib/nanoc/core/utils.rb
@@ -14,12 +14,24 @@ module Nanoc
       def self.expand_path_without_drive_identifier(file_name, dir_string)
         res = File.expand_path(file_name, dir_string)
 
-        if Nanoc::Core.on_windows?
+        if windows_fs?
           # On Windows, strip the drive identifier, e.g. `C:`.
           res = res.sub(/^[A-Z]:/, '')
         end
 
         res
+      end
+
+      # Returns `true` if absolute file paths start with a drive identifier, like `C:`.
+      def self.windows_fs?
+        # NOTE: This isn’t memoized with ||= because @_windows_fs is a boolean.
+
+        return @_windows_fs if defined?(@_windows_fs)
+
+        absolute_path = File.expand_path('/a')
+        @_windows_fs = absolute_path.start_with?(/^[A-Z]:/)
+
+        @_windows_fs
       end
     end
   end


nanoc-core/lib/nanoc/core/utils.rb
```diff
@@ -14,13 +14,21 @@ module Utils
...
           # On Windows, strip the drive identifier, e.g. `C:`.
           res = res.sub(/^[A-Z]:/, '')
         end
 
         res
       end
+
+      # Returns `true` if absolute file paths start with a drive identifier, like `C:`.
+      def self.windows_fs?
+        @_windows_fs ||= begin
```
Fjan, line 27: On Mac / Linux this will usually return `false`, so for the memoization to work it should probably be something like `@_windows_fs = ...   if @_windows_fs.nil?`

denisdefreyne: Oh! Good catch --- will fix.

(I have been writing an article on memoization in Ruby. The *irony*.)

nanoc-core/lib/nanoc/core/utils.rb
```diff
@@ -14,13 +14,25 @@ module Utils
...
         end
 
         res
       end
+
+      # Returns `true` if absolute file paths start with a drive identifier, like `C:`.
+      def self.windows_fs?
+        # NOTE: This isn’t memoized with ||= because @_windows_fs is a boolean.
+
+        return @_windows_fs if defined?(@_windows_fs)
```
Fjan, line 29: Ok, great, I think this works correctly. Just to get pedantic, because this is a module function it might be better to memoize with class variable `@@_windows_fs` so it only gets evaluated once instead of for every instance. And because the line before last line already evaluates to the return value it's possible to leave off the very last line. But feel free to ignore. :-)

denisdefreyne: I’ve opted for a “class instance” variable, which will work as intended because the method already is a class/module method. Class variables in Ruby have odd behavior; see the [Ruby style guide entry on class variables](https://github.com/rubocop/ruby-style-guide?tab=readme-ov-file#no-class-vars).

I can indeed leave off the last line, but I like the explicitness of it. (And Rubocop does not complain about it, so I’m happy with it.)

denisdefreyne:
This is now released in Nanoc 4.13.2.

Fjan:
Yes, I can confirm it works on a PC now. Thanks!

PR #1727, opened 2025-02-15
Make clonefile a hard dependency

denisdefreyne:
### Detailed description

`clonefile`, which is a good idea in general, should now be well supported across all operating systems. The compatibility issues should be a thing of the past.

### To do

* [x] Tests

### See also

- #1511 
- https://codeberg.org/da/ruby-clonefile

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -56,7 +56,6 @@ group :plugins do
   gem 'asciidoctor'
   gem 'bluecloth', platforms: :ruby
   gem 'builder'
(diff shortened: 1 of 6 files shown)


denisdefreyne:
Test failure is unrelated; will fix on `main`.

denisdefreyne:
FYI @da2x — *finally* 😆

denisdefreyne:
This is now released in Nanoc 4.13.4.

PR #1734, opened 2025-03-09
Run nanoc checks in parallel

Fjan:
### Detailed description

This speeds up nanoc checks by running them in parallel. We've used this on a large nanoc site with dozens of tests for over a year now, saving hours. If there is a single check it keeps the existing behaviour, because it allows for better feedback in case of a longrunning check (and that way no tests needed to be modified).

I used the parallel gem, because some parts of the check runner are not thread safe. This required me to replace the class that is passed into `Nanoc::Checking::Issue` with a Struct to allow it to serialize. It might be a bit cleaner to rewrite the Issue class to take a symbol rather than a class, but that would increase the scope of this PR further than I was comfortable with.

### To do

* [ ] Tests? The existing tests pass, I don't know how to test this better.
* [ ] Documentation
* [ ] Feature flag? If someone has a nanoc check that modifies state, like reformatting something in your site, then running the checks in parallel could cause undefined behaviour. I would consider this very bad practice, not sure if we need to take it into account.

### Related issues

https://github.com/nanoc/features/issues/47

### DIFF
--- a/nanoc-checking/lib/nanoc/checking/runner.rb
+++ b/nanoc-checking/lib/nanoc/checking/runner.rb
@@ -87,29 +87,28 @@ module Nanoc
         end
       end
 
+      ClassID = Struct.new(:identifier)
       def run_checks(classes)
         return [] if classes.empty?
 
         # TODO: remove me
         Nanoc::Core::Compiler.new_for(@site).run_until_reps_built
-
-        checks = []
-        issues = Set.new
         length = classes.map { |c| c.identifier.to_s.length }.max + 18
-        classes.each do |klass|
-          print format("  %-#{length}s", "Running check #{klass.identifier}… ")
-
+        if classes.size==1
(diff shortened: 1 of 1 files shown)


denisdefreyne:
Thanks for the idea! This is definitely helpful, though I’ve got an improvement in mind that’ll make the output more clear. I’ll follow up soon.

denisdefreyne:
Here is an idea of what I’m thinking about:

https://github.com/user-attachments/assets/d95860d3-3473-4da1-9b6d-b2f7870bd36d

denisdefreyne:
See #1735 for an alternate implementation based on yours. (Still WIP for now.)

Fjan:
Nice! I agree, your implementation is much better. Feel free to discard mine.

denisdefreyne:
Closing in favor of #1735.

PR #1735, opened 2025-03-15
Run checks in parallel

denisdefreyne:
### Detailed description

Similar to #1734: runs checks in parallel. This is what it looks like:

https://github.com/user-attachments/assets/aa9af20c-187a-425a-8ce0-ef07a00f3c70

Parallelism is currently set to 5, but is configurable. (I think it’s good to have a limit, but I’m not sure what it should be.)

I can’t imagine individual checks sharing state, so parallelising this should be possible without side effects. For that reason, like in ##1734, I think a feature flag wouldn’t be useful.

### To do

* [x] Tests
* [ ] ~~Documentation~~
* [ ] ~~Feature flags~~

### Related issues

See #1734.

### DIFF
--- a/nanoc-checking/lib/nanoc/checking/runner.rb
+++ b/nanoc-checking/lib/nanoc/checking/runner.rb
@@ -6,9 +6,14 @@ module Nanoc
     #
     # @api private
     class Runner
+      # Number of threads to use for running checks.
+      NUM_THREADS = 5
+
       # @param [Nanoc::Core::Site] site The Nanoc site this runner is for
       def initialize(site)
         @site = site
+
+        @log_mutex = Thread::Mutex.new
       end
 
       def any_enabled_checks?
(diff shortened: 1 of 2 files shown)


denisdefreyne:
@Fjan FYI! Can you give this a try and see how this works for you?

nanoc-checking/lib/nanoc/checking/runner.rb
```diff
@@ -93,25 +98,74 @@ def run_checks(classes)
...
+        $stdout << "\e[#{classes.count}A" # move to top
+
+        # Print all checks (all “pending” for now)
+        puts 'Running checks…'
+        checks.each_with_index do |check, index|
+          log_check(index:, topic: format("  %-#{length}s", check.class.identifier.to_s), state: 'pending')
+        end
+
+        # Run checks in parallel
+        indexes = (0..checks.count - 1).to_a
```
Fjan, line 115: `indexes = checks.count.times.to_a` is perhaps slightly prettier

Fjan:
> @Fjan FYI! Can you give this a try and see how this works for you?

Tried on our largish code base and it works great!

Fjan:
Oh, cheered too early, I noticed on a nanoc site with 6 parallel tests only the first 5 get run

nanoc-checking/lib/nanoc/checking/runner.rb
```diff
@@ -93,25 +98,74 @@ def run_checks(classes)
...
+
+        # Collect issues
+        issues = Set.new
+        checks.each do |check|
           issues.merge(check.issues)
+        end
 
-          # TODO: report progress
+        # Move cursor to below list
+        $stdout << "\e[#{checks.count}B"
```
Fjan, line 134: This needs one extra \n to prevent the output from the last test from being overwritten: 
`$stdout << "\e[#{checks.count}B\n"`

denisdefreyne: Good catch! It’s in fact the `puts 'Running checks…'` that’s throwing sand in the works. Moving that earlier fixes the issue.

nanoc-checking/lib/nanoc/checking/runner.rb
denisdefreyne: Oh yes, much cleaner with `Parallel.each_with_index`.

Fjan:
Great! I have a small tangentially related PR I would like to submit on top of this, but I can't seem to get the master branch to run, something about the AnsiStringColorizer changing from a module to a class. I think that happens because the master branch refers to the locally installed older nanoc. I'll wait for a release, unless this rings a bell and you know how to get around that.

denisdefreyne:
> I can't seem to get the master branch to run, something about the AnsiStringColorizer changing from a module to a class

Odd! Can you double-check that you don’t have any stray files that define `ANSIStringColorizer` anywhere?

Fjan:
Actually, I didn't realize you have to specify the path to each part of nanoc in your bundle separately, if you just specify the main 'nanoc' then it will still pick up the installed gems for the other parts. (Thanks ChatGPT!)
```ruby
gem 'nanoc', path: '../nanoc'
gem 'nanoc-core', path: '../nanoc/nanoc-core'
gem 'nanoc-cli', path: '../nanoc/nanoc-cli'
gem 'nanoc-checking', path: '../nanoc/nanoc-checking'
gem 'nanoc-deploying', path: '../nanoc/nanoc-deploying'
```

Fjan:
(BTW, that small PR I mentioned that could go on top of this is here: https://github.com/nanoc/nanoc/pull/1739/files let me know if you would like me to create a separate issue for it)

denisdefreyne:
This is now released in Nanoc 4.13.4 and `nanoc-checking` 1.0.6.

PR #1739, opened 2025-03-17
Provide helper to check only changed files

Fjan:
This PR provides a helper method that filters `output_filenames` into `output_changed_filenames`, containing only the files that changed since the last successful run.

### Detailed description

Many checks, such as HTML5 validation and spelling checks, can be greatly sped up by skipping all files that have been checked before and have not changed since then. Instead of each check having to do it's own bookkeeping this helper method provides the list of filenames that changed since the last time the check completed without any issues.

Tests: I've not written rspecs before, but if this PR is acceptable I can see if ChatGPT can come up with one
Documentation: One potential pitfall is that while you are modifying a check you can get confused if you've completed a run successfully, and then subsequently running the modified check will not show anything. We could invalidate the cache if a check is modified but that seems a lot of work for just this edge case.

### To do

* [ ] Tests
* [ ] Documentation

### Related issues

### DIFF
--- a/nanoc-checking/lib/nanoc/checking/check.rb
+++ b/nanoc-checking/lib/nanoc/checking/check.rb
@@ -42,6 +42,8 @@ module Nanoc
             reps:,
           )
 
+        @store = Nanoc::Core::Store.tmp_path_for(config: site.config, store_name: 'last_good_check')
+
         context = {
           items: Nanoc::Core::PostCompileItemCollectionView.new(site.items, view_context),
           layouts: Nanoc::Core::LayoutCollectionView.new(site.layouts, view_context),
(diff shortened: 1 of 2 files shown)


denisdefreyne:
Hey Jan! I believe the functionality in this PR would cause confusion and as such I think it needs some adjustments to work well.

My primary concern is that it would not at all be obvious that some checks are, in fact, failing, if those checks use `#output_changed_filenames`. Additionally, the decision of whether to run a check on *all* files or just changed files is made in the check implementation itself, rather than by the user, which I think needs to change too.

While I do think it’s useful to have functionality for narrowing/focusing the checks to recently updated files, I think a better user experience could be achieved by allowing a filter to be specified, e.g. `nanoc check --updated-since=2025-04-20T09:39:24+0200`. The `#output_changed_filenames` would then return only the output filenames that have a mtime more recent than this one.

While `nanoc check` itself wouldn’t have functionality to track when the last check was run, it would be scriptable, like so:

```bash
nanoc check --updated-since=`ruby -e 'print File.mtime("README.md").iso8601'`
touch tmp/last_check_ts
```

What do you think?

Fjan:
Hi Denis! It should indeed never be possible for a check to pass when it only checks the changed files, if it would fail if it checks all files, that would be too confusing. But this PR addresses that concern. This is why I deliberately put the choice to limit the check to changed files one the designer of the check: Some checks are never safe to run on just the changed subset. For example the `internal_links` check cannot use it because a change in one file can break the referring link to it in another unchanged file, so that check will not use `#output_changed_filenames`. The users (at least my users 😉) will typically not realize such subtleties. For that reason I don't think I would want to offer a general `updated-since` option to my users because it could tell them the checks pass when a full check would fail. 

Secondly, if a check can be safely run on a subset, such as an HTML5 validator check or a spell check, then one of the checks may complete successfully while the other one fails. Storing the last known good run for each check ensures, for example, the HTML5 validation does not have to run again when there is just a small typo in one file. (And we have been using this functionality as a monkey patch for quite some time with many people so I'm pretty confident it works as advertised).

PR #1743, opened 2025-04-20
remove haml warnings related to :outvar and :disable_capture which are now invalid

doc75:
### Detailed description

When using the Haml template format we have now the following warnings:
```
Haml::TempleEngine: Option :disable_capture is invalid
Haml::TempleEngine: Option :outvar is invalid
```
This PR is fixing this by removing those options from the Haml filter

### To do

I have tested this change on one of my site with success.
I have not found any test that would need to be modified for this (but I might have missed it).
I do not think that any documentation is necessary, but let me know if I am wrong.

* [ ] Tests
* [ ] Documentation

### Related issues

None

### DIFF
--- a/nanoc/lib/nanoc/filters/haml.rb
+++ b/nanoc/lib/nanoc/filters/haml.rb
@@ -17,8 +17,6 @@ module Nanoc::Filters
       # Get options
       options = params.merge(
         filename:,
-        outvar: '_erbout',
-        disable_capture: true,
       )
 
       # Create context


denisdefreyne:
@doc75 I have not been able to reproduce this issue. `disable_capture` and `outvar` are both accepted.

Could it be that you are running an old version of Haml? `Haml::TempleEngine` suggests that you’re using Haml 5.x or older; the current version of Haml is 7.0.

doc75:
@denisdefreyne actually, when checking the code, it seems that the support for Haml 7 is done and it is just a question of time to have it released. So I will not open an issue and just be patient (no rush on my side).

denisdefreyne:
This is now released in Nanoc v4.14.4. Thanks!

PR #1745, opened 2025-04-21
Remove clonefile as a hard dependency

### LINKED ISSUE #1744: Nanoc 4.13.4 can no longer be installed on Windows
In https://github.com/nanoc/nanoc/pull/1727 the clonefile gem was added as a hard dependency. However, the clonefile gem can only be installed on Linux and Mac, not on Windows, so trying to install nanoc now fails on Windows. 

Bundler supports installing gems everywhere except windows via the `platforms` parameter, like this:
`gem 'clonefile', '~> 0.5.2', platforms: :ruby`
I suggest we roll back the PR and replace it with that line instead.

denisdefreyne:
[clonefile](https://codeberg.org/da/ruby-clonefile) unfortunately cannot be installed on Windows, and Windows support is needed.

This reverts commit 982575af3edc26e4a87b40c721da6104a4cbaaa1, but adds clonefile to the Gemfile.

The clonefile gem can still be installed, but it can’t be a hard dependency.

Fixes #1744.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,6 +22,7 @@ end
 
 group :devel do
   gem 'addressable', '~> 2.8'
+  gem 'clonefile', '~> 0.5.3', platforms: :ruby
   gem 'contracts', '~> 0.17'
   gem 'debug', '~> 1.9'
   gem 'fuubar', '~> 2.5'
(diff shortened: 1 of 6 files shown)


denisdefreyne:
@da2x Is there a different way to make clonefile still work on Windows? Or be a no-op or so?

I’ll likely merge and release this revert, just to make Windows users able to use the most recent Nanoc release.

